PR #140, opened 2026-08-07
fixing chart overflow on xl screens

aksperiod:
Resize bug is mostly fixed, thanks for that! on my monitor it is still broken, but it is no longer a chart.js issue and now is just because `SummaryData` is overflowing at the `xl` breakpoint. Picture attached, where you can see the chart overflows.

Screenshot 2026-08-07 at 9 56 23 AM


mle17 (approved):
Feel free to merge. The image baselines can be updated in a separate PR to fix GitHub's pipeline builds.

PR #173, opened 2026-08-15
feat(pipeline): keep stop identity and alightings at ingest

mle17:
Wave 1 of a 5-PR batch. Owns the frozen contract PRs 2–5 read. **No data change** — this is the refactor and the new aggregation only.

`data/raw/` (four committed zips of CPRA Excel exports) is already stop/station-level with ons, offs and activity for all three day types. `aggregate_to_line_ridership` did `groupby("LINE").sum()` and threw stop identity, offs and activity away at ingest. This stops discarding it.

Roadmap for the whole batch: [`docs/ROADMAP.md`](docs/ROADMAP.md).

## What changed

- **`extract_leaf_rows(df, mode)`** — the leaf-row rule (bus `DIRECTION notna & != "Total"`; rail `STATION_ORDER notna & != "Total"`) **and** the ROUTE-over-LINE resolution, in one place. Both aggregations wrap it, so the rule cannot drift between them.
- **`aggregate_to_line_ridership`** — signature and output unchanged, now a thin wrapper.
- **`aggregate_to_stop_ridership(df, year, month, mode)`** — one row per line per stop. `*_OFFS` kept (new signal the app has never had), `*_ACT` dropped (equals ons+offs), bus direction collapsed, raw decimals so rounding stays on write.
- **`scripts/stop_identity.py`** — `normalise_stop_name`, `display_stop_name`, `strip_rail_platform_suffix`, `parse_station_order`, `stop_key`. Pandas-free.
- **`scripts/stop_aliases.json`** — hand-editable rename table, ships empty.
- **Docs** — `scripts/README.md` gains `stop_identity`, the alias table with a `shakeups.json`-style provenance section, and the reconciliation caveat. `docs/ROADMAP.md` is new.

## The gate

`src/data/ridership.json` must be unchanged after a full re-ingest of all four archives.

**It is — but the baseline is not what it looks like.** A control run on `main` **before any of my edits** already produces a 40-line diff: `fill_missing_months` synthesises zero rows for line 106 in 2026-01 … 2026-05 that are not in the committed file. So the committed `ridership.json` is not the fixed point of a full re-ingest, at `main`, today.

Judged against that control:

```
ridership.json:                  BYTE-IDENTICAL to control (sha256 f03df3a9…)
metro_line_metadata_current.json: BYTE-IDENTICAL to control
```

Both files were reverted afterwards; **nothing under `src/` is in this diff**.

The line-106 drift is a real gap and it is **not mine to patch** — `process_ridership.py` belongs to a later PR in this batch. Flagging it rather than fixing it. It is either five months of genuinely-missing line 106 data or a stale commit of `ridership.json`; someone should decide which before PR 3 lands, because PR 3 is the one that rewrites these files for real.

## Verification

```
pytest scripts/  →  285 passed
```

Six existing test files unmodified. `test_convert_excel_ridership.py` extended: **254 insertions, 0 deletions** — every existing assertion untouched, and the new stop tests reuse `_make_bus_df_with_totals` / `_make_rail_df_with_totals` verbatim.

Required tests present:

- **D Line not attributed to the B Line at stop grain** — `TestDLineNotAttributedToBLine`, three assertions: 805's stations are filed under 805, 802's set does not contain them, and 802's boardings exclude 805's.
- **Reconciliation invariant** — `TestReconciliationInvariant`, six frames × three day types: per-line sums of `aggregate_to_stop_ridership` equal `aggregate_to_line_ridership`'s `Riders`, asserted **pre-rounding**.

Against the real archives (not just fixtures), rail reconciles to 0.0 and bus to 3.6e-12:

```
Rail 2025-12: 124 rows, 107 keys, DX/SA/SU max abs diff 0.0
              805 → 8 stations, 802 → 14, Union Station one key across 3 lines
Bus  2025-12: 8,953 rows, 6,775 keys, DX max abs diff 3.6e-12
              0 non-URL-safe keys, 0 keys with more than one display name
```

## Decisions I made rather than stopping to ask

- **`normalise_stop_name` case-folds, so it cannot be a display string.** Added `display_stop_name(mode, raw)` for the reader-facing form. Two clearly-named functions beat one ambiguous one, and rows sharing a `stop_key` must also share a name or the UI shows one place under two labels. Additive to the plan's signature list.
- **`stop_key(mode, name, aliases=None)`** — the plan's three positional params, with a default so the function stays pure for tests. `aggregate_to_stop_ridership` keeps exactly the four params the plan specifies and loads the alias table itself, so there is one source of aliases; a test wanting a custom table monkeypatches `stop_identity.load_aliases`.
- **Alias chains are followed, cycles raise.** A stop renamed twice would otherwise still split.
- **`stop_aliases.json` ships empty**, and that is checked, not assumed. Across all four archives, no month has both an added and a dropped key — which is exactly the rename signature the §1.3 guard looks for:

  ```
  Bus  2025-07 … 2026-05:  +40/-0 (2025-12), +0/-8 (2026-01), otherwise flat
  Rail 2025-07 … 2026-05:  +6/-0 (2025-09, route breakout), +3/-0 (2026-05, D Line extension)
  ```

  The known rail renames (`APU Station` → `APU / Citrus College Station`) are GTFS-side, not ridership-side, so they belong to PR 2's geometry join.
- **Rail platform suffixes** are stripped by a regex covering the 6 forms that occur in the data (`- A Line`, `- A / E Line`, `- Metro Red & Purple Lines`, …), table-driven over all 12 distinct real names. It is **rail-only**: bus names like `El Monte Station - Upper Level` and `Del Amo Station - Discharge Only` are genuinely distinct stops.
- **`extract_leaf_rows` coerces all six `LEAF_VALUE_COLS`**, not just the three ONS columns the line path reads. Harmless to the line output (it selects only ONS) and it means the stop path does not restate the coercion.

## Not in scope, deliberately

- **The §1.3 rename guard is not implemented.** It compares a month against months already stored, so it belongs with `detect_renames` in `stop_ridership.py` — PR 3's file, not one I own. The identity primitives it needs are here. Noted in `scripts/README.md` and `docs/ROADMAP.md`.
- **`CONTEXT.md` is untouched.** **Boardings** / **Alightings** (avoid: ons, offs), Stop Place, Stop Ridership Record, Stop Readout, Stop Coverage Window are not registered — `CONTEXT.md` is not in this PR's file scope and PRs 2–5 run in parallel against it. `scripts/README.md` states the vocabulary rule inline meanwhile. **PR E owes the CONTEXT.md entries.**
- **`docs/README.md` gains one line** indexing `docs/ROADMAP.md`. That is one file outside the scope I was given, added after review: the repo's own rule is that `docs/README.md` lists every document and "everything else is one link from there", so adding an unindexed doc would itself have been the defect. Nothing else in that file moved.

## Review

Two independent blind reviews, both against the real archives rather than the fixtures. Round 1 raised three findings, all fixed in `7bd4736`: the `STATION_ORDER` rationale cited the wrong event (112 → 124 is the A Line Foothill extension plus the ROUTE 805 breakout at 2025-09, not the D Line extension, which is 124 → 127 at 2026-05, and no existing number shifted in any month); `normalise_stop_name(float("nan"))` returned `"nan"`, so a blank `STOP_NAME` would have become a plausible stop `bus:nan` carrying every blank-named row on its line; and `ROADMAP.md` was unindexed.

Round 2 blocked nothing. Its numeric findings are fixed in `ef699a4`, each re-measured before being written:

- the quoted `p25 = 4.4` was leaf grain for one month in a stop-grain paragraph — stop grain is p25 8.9, 93% non-integral
- **the plan's `< 0.02` reconciliation tolerance is already breached by the data** — over 1,252 line-months, median 0.04% / p95 0.49% / max 2.35%, with three at or above 2%. PR 3's `--check` needs a looser bound. And in this window `Days = 1` is hardcoded, so the days-weighted average contributes *nothing* — the whole spread is per-stop rounding.
- **the §1.3 rename guard, as specified, would miss the one relocation-shaped event in this data.** Line 28's `bus:san-vicente-fairfax` (2025-07 → 2025-12) and `bus:san-vicente-orange-grove` (2025-12 → 2026-05) overlap by a month, so the add lands in 2025-12 and the drop in 2026-01 and a same-month rule sees neither. That pair is also left un-aliased on purpose — rename or two-block move, the ridership data cannot distinguish them, and aliasing on a guess merges two real series.
- 2025-12 was the month omitted from the payload figures, and it is the outlier (15,146 → 8,953)

Both are recorded in `docs/ROADMAP.md` so PR 3 inherits them rather than rediscovering them.

## Files

Created: `scripts/stop_identity.py`, `scripts/stop_aliases.json`, `scripts/test_stop_identity.py`, `docs/ROADMAP.md`
Modified: `scripts/convert_excel_ridership.py`, `scripts/test_convert_excel_ridership.py`, `scripts/README.md`

Nothing under `src/`, `vite/`, `e2e/`, `public/`, `data/`, and no other `scripts/*.py`.


mle17:
## Review — PR 1 of the stop-level batch

Reviewed against `16f0067..origin/feat/stop-level-ingest` (head `9382d31`), the approved plan's
Part 1 §1.1–§1.4 / §1.7, `CLAUDE.md`, `CONTEXT.md`, `docs/how-it-works.md`, ADR-0001, ADR-0007,
and both versions of `scripts/README.md`. Everything below was executed, not inferred.

**Verdict: the code is correct and the must-not-drift checklist passes in full.** Three findings,
all documentation or latent robustness. None blocks correctness. One (F1) I'd want fixed before
this lands, because it is the stated justification for a decision PRs 2–5 inherit. Merge call is
Michael's.

---

### Must-not-drift checklist — all verified

**1. `aggregate_to_line_ridership` is unchanged.** Loaded both versions side by side (base via
`git show 16f0067:scripts/convert_excel_ridership.py`) and ran them over all **22 real frames**
extracted from the four archives in `data/raw/` — 11 bus months, 11 rail months, 2025-07 → 2026-05.
`pd.testing.assert_frame_equal(..., check_exact=True, check_dtype=True)` on every pair:

```
22 real frames loaded
line-level identical on 22/22 frames
```

The refactor does widen the numeric coercion from the 3 `*_ONS` columns to all 6 `LEAF_VALUE_COLS`.
That cannot reach the line output — the line aggregation only reads `WD_ONS`/`SA_ONS`/`SU_ONS` — and
the 22/22 result confirms it. The bus branch also gains a `.copy()` that the base version lacked,
which removes a latent `SettingWithCopyWarning` without changing values.

**2. The leaf-row rule exists in exactly one place.** `rg` for the bus and rail filters across the
whole repo returns two lines, both inside `extract_leaf_rows`:

```
scripts/convert_excel_ridership.py:139:  leaf = df[df["DIRECTION"].notna() & (df["DIRECTION"] != "Total")].copy()
scripts/convert_excel_ridership.py:144:  leaf = df[df["STATION_ORDER"].notna() & (df["STATION_ORDER"] != "Total")].copy()
```

No other occurrence anywhere — not in `process_ridership.py`, not in `update_ridership.py`, not in
the tests.

**3. ROUTE-over-LINE survives, at both grains, on the real archives.** Per rail month, at stop grain:

| Month | Lines | 802 stops | 805 stops |
|---|---|---|---|
| 2025-07 / 08 | 801 802 803 804 807 | 14 | 0 (805 not yet reported separately) |
| 2025-09 → 2026-04 | 801 802 803 804 805 807 | 14 | 8 |
| 2026-05 | 801 802 803 804 805 807 | 14 | 11 |

802 holds exactly the 14 B Line stations; 805 holds the 8 D Line stations, going to 11 in 2026-05
when Wilshire/La Brea, Wilshire/Fairfax and Wilshire/La Cienega appear. The six stops that appear
under both are the shared Union Station → Wilshire/Vermont trunk, which is correct — both routes
call there and each reports its own boardings. Wilshire/Normandie and Wilshire/Western never appear
under 802.

**4. `station_order` is never identity.** The groupby key is `["LINE", "stop_key"]`;
`station_order` is carried as an aggregated attribute (`min`) and appears in no key, no join, no
sort tiebreak. It is rail-only (`Int64`, all-NA for bus). Confirmed empirically that it *is* per-route
and therefore could not be one: Union Station is `1026` on 801, `4001` on 802, `5001` on 805.

**5. No file outside the allowed list is touched.**

```
docs/ROADMAP.md                        130  0
scripts/README.md                      114  0
scripts/convert_excel_ridership.py     107 18
scripts/stop_aliases.json               19  0
scripts/stop_identity.py               195  0
scripts/test_convert_excel_ridership.py 254 0
scripts/test_stop_identity.py          296  0
```

Nothing under `src/`, `vite/`, `e2e/`, `public/`, `data/`; no other `scripts/*.py`.

---

### The other required gates

**Re-ingest of all four archives — report this one carefully.** The acceptance gate as written
("`git diff --stat src/data/ridership.json` empty after a full re-ingest") **is not met — and it is
not met at the base commit either.** I established the control first.

- Control, base converter → `process_ridership.py`: `ridership updated: 42,633 -> 42,638 (+5)`,
  `sha256(ridership.json) = f03df3a9…`
- `git checkout src/data/`, then the PR's own CLI over all four zips:
  `ridership updated: 42,633 -> 42,638 (+5)`, `sha256(ridership.json) = f03df3a9…`

**Byte-identical.** The +5 rows are `line_name: 106` at 2026-01…05 with `0.0` for all three day
types — a pre-existing quirk of the append-only merge, present at `16f0067`, untouched by this PR.
`metro_line_metadata_current.json` unchanged in both runs. `src/data/` restored with
`git checkout src/data/`; tree verified clean (`bcd29ba7…`, the committed hash).

So the correct statement of the gate is **"a full re-ingest produces exactly what it produced at the
base commit"**, and that holds byte for byte. The gate's literal wording was never satisfiable; that
is a bug in the plan, not in the PR. Worth correcting in `docs/ROADMAP.md`'s gate column so PR 3
doesn't inherit an impossible check.

**`pytest scripts/` green.** 282 passed, from both `scripts/` and the repo root. The six pre-existing
test files: five have zero diff; `test_convert_excel_ridership.py` is `254 / 0` — **254 insertions,
zero deletions**. Extended, not modified. Test count 36 → 62 there, plus 42 new in
`test_stop_identity.py`.

**D Line not attributed to the B Line at stop grain.** `TestDLineNotAttributedToBLine`, three
assertions — station sets under 805, station sets under 802, and boardings split 300/100. Backed by
the real-archive check in item 3 above.

**Reconciliation invariant, pre-rounding.** `TestReconciliationInvariant` covers six frames. I also
ran it over all 22 real frames × 3 day types:

```
max abs per-line discrepancy across all real frames: 3.6e-12
```

Float noise only.

**`_make_bus_df_with_totals` / `_make_rail_df_with_totals` reused verbatim** — unmodified, and both
are used by `TestExtractLeafRows`, the exclusion tests and the reconciliation tests.

**`*_ACT` dropped, `*_OFFS` kept** — absent from `STOP_OUTPUT_COLS`, asserted by
`test_activity_columns_dropped`, and `test_alightings_preserved` pins the offs.

**Vocabulary.** `scripts/README.md` states the Boardings/Alightings rule explicitly and the pipeline
columns stay `*_ONS`/`*_OFFS`. No UI copy in this PR.

---

### Findings

#### F1 — the `STATION_ORDER` justification does not match the archives *(docs; fix before merge)*

`scripts/stop_identity.py` (module docstring), `scripts/README.md` and
`aggregate_to_stop_ridership`'s docstring all assert:

> rail leaf rows went 112 → 124 between 2025-07 and 2025-12 as the D Line extension landed, moving
> every downstream number

Measured against `data/raw/`, each clause is wrong:

- Rail leaf rows per month: `112, 112, 124, 124, 124, 124, 124, 124, 124, 124, 127`. The 112 → 124
  step is at **2025-09**, and it is the **A Line Foothill extension** (Glendora, San Dimas,
  La Verne/Fairplex, Pomona North) **plus ROUTE 805 first being reported as its own route** (8
  stations) — 4 + 8 = 12.
- The **D Line extension is the 124 → 127 step at 2026-05** (Wilshire/La Brea, Fairfax, La Cienega).
- **No downstream number moved.** Diffing `station_order` per `(line, stop_key)` against 2025-07:
  `0 changes` in every one of the other 10 months. The new stations were appended (801: `1045–1048`
  after `1044`), not inserted.

The *decision* is still right, and the real evidence is stronger than the cited one: `station_order`
is per-route, so Union Station carries three different numbers in the same month (`1026`/`4001`/
`5001`) and cannot be an identity under any circumstances. But `CLAUDE.md`'s docs rule and
`docs/README.md`'s authority order both make a doc that states something the data does not a defect,
and this text is the rationale PRs 2–5 will cite. Same correction needed in the
`aggregate_to_stop_ridership` groupby comment, which gives "a station's sequence number moves when
the route is extended" as the reason for `min` — a reason not observed in this window.

#### F2 — a NaN `STOP_NAME` silently becomes the stop `bus:nan` *(latent; not triggered today)*

`normalise_stop_name`/`display_stop_name` guard only `raw is None`, while `parse_station_order` in
the same module explicitly handles NaN with the `value != value` trick. So:

```
stop_key("Bus", None)          -> ValueError: no usable characters   # loud
stop_key("Bus", float("nan"))  -> "bus:nan"                          # silent
```

With two blank-named leaf rows on line 90 the frame comes back with a real stop named `nan` whose
boardings are the sum of every blank-named row on that line. Not reachable from today's data — I
checked all 11 bus months and found **0 leaf rows with a null or blank `STOP_NAME`** — but it is
exactly the kind of thing a changed export brings, and it fails silently rather than loudly, which
is the opposite of how the rest of this module behaves. PR 3's rename guard would eventually notice
a new key, by which time the merged number is already in the file. A `value != value` check in
`normalise_stop_name` would make it raise like the `None` case.

#### F3 — `docs/ROADMAP.md` is orphaned *(docs; trivial)*

Nothing in the repo links to it — `rg ROADMAP` over all Markdown returns zero hits outside the file
itself. `docs/README.md` opens with "Every document in this repo, what it's for, and when you'd read
it" and `CLAUDE.md` says "docs/README.md — the index. Everything else is one link from there." One
row in the Reference table closes it.

---

### Notes, not findings

- **Slug collisions are real but currently benign.** Normalisation merges 3 bus `(line, stop)` groups
  per month that exact-name grouping keeps apart: `bus:pacific-rr-xing` (`Pacific / RR-- Xing` +
  `Pacific / RR-Xing`) and `bus:district-produce-plaza`. All three are typo variants of one stop, so
  the merge is a win — but the mechanism would merge two genuinely different places just as quietly.
  PR 2's `spread_m` warning is the natural place to catch the geographic case. Side effect: the
  plan's "14,927 → 8,841 rows/month" is **8,838** in practice; `docs/ROADMAP.md` repeats 8,841.
- **Rail suffix stripping over-merges nothing.** All 12 suffixed names in the archives are covered by
  the table-driven test, and `(line, stop_key)` group sizes are `1` for every rail month — stripping
  never collapses two distinct rows within one route. 116 raw names → 110 keys, exactly the 6
  two-platform stations.
- **Display names are stable across months.** `stop_name` is picked by `min` within a frame; across
  all 11 bus months, **0 keys** vary. Worth keeping an eye on in PR 3 — a month where only one spelling
  variant appears could flip the label, though `min` keeps it deterministic per run.
- Empty leaf frames are handled: both aggregations return correctly-shaped zero-row frames.
- `aggregate_to_stop_ridership` does not mutate the caller's frame; `test_returns_a_copy` pins it.
- `_resolve_alias`'s over-length error interpolates the reassigned `slug`, so it names the last hop
  rather than the chain's origin. Cosmetic, and only reachable on a genuine 10-hop chain since the
  cycle check fires first.
- `CONTEXT.md` does not yet register Boardings/Alightings, though `scripts/README.md` now states the
  rule. `docs/ROADMAP.md` assigns that to PR 5, which is a defensible reading of the sequencing even
  though `docs/agents/domain.md` would have put it first.

---

Checks run: 22-frame base-vs-PR frame equality; repo-wide leaf-rule grep; 802/805 separation at both
grains on real archives; `station_order` drift across all 11 months; controlled re-ingest at base and
at PR head with `git checkout src/data/` between and after; `pytest scripts/` from two directories;
reconciliation over 22 frames × 3 day types; NaN / empty-frame / slug-collision / suffix-over-strip
probes. Working tree left clean.

mle17:
**Second-round review** — independent pass over `16f0067..feat/stop-level-ingest` (head `7bd4736`), formed without reading the first round. Everything below was executed, not inferred.

**Nothing blocks.** Eight non-blocking items, four of them forward-looking for PR 3. Merge authority is Michael's.

---

## Must-not-drift checklist — all six verified

**1. `aggregate_to_line_ridership` is unchanged, on real data.** Wrote `git show 16f0067:scripts/convert_excel_ridership.py` to a scratch module, imported both side by side, and ran every inner workbook in all four archives through each — 22 frames, 11 months × 2 modes. `pd.testing.assert_frame_equal(..., check_exact=True)` → **0 mismatches**.

**2. The leaf-row rule exists in exactly one place.** Repo-wide grep for the two filters: `convert_excel_ridership.py:139` (bus `DIRECTION`) and `:144` (rail `STATION_ORDER`), both inside `extract_leaf_rows`. No other occurrence in `scripts/`, `src/`, `vite/`, `e2e/`.

**3. ROUTE-over-LINE survives, at both grains, against the archives.**

```
          B(802) stops   D(805) stops   B lineRiders   B stopSum
2025-07        14              0            59417.0     59417.0
2025-09        14              8            48194.0     48194.0
2026-05        14             11            47611.0     47611.0
```

802 holds 14 stations in all eleven months; 805 breaks out at 2025-09 with 8 and reaches 11 at 2026-05. `rail:wilshire-normandie-station` and `rail:wilshire-western-station` are on 805 only, never on 802, and 802's per-stop boardings sum exactly to its line-level `Riders` in every month.

**4. `station_order` is never identity.** It is produced, `min`-aggregated, and emitted. It appears in no `groupby` key, no merge, no dict key. Independently: across the eleven months, **0** `(line, stop_key)` pairs carry more than one `station_order`, and **0** groups carry more than one within a single frame — so `min` is never actually choosing between values today.

**5. File scope.** The seven listed files, plus the one-line index entry in `docs/README.md`. That line is **required**, not scope creep — `docs/README.md` is the index and the repo's own rule is that everything else is one link from there, so adding `ROADMAP.md` without indexing it would have been the defect. (The PR body still says the opposite; see F3.)

**6. The re-ingest gate.** Ran the full four-archive re-ingest at the PR head, `git checkout src/data/`, then re-ran it with the base `convert_excel_ridership.py` from a scratch scripts directory:

```
committed          bcd29ba72c1d247d5a9b336da02b23831ea90c2a7e7c508b75f6345b1348f17a
PR-head re-ingest  f03df3a92d7281dde2165e928a1ecdd87464cef4cd0155cf5d139e84e61a2123
base re-ingest     f03df3a92d7281dde2165e928a1ecdd87464cef4cd0155cf5d139e84e61a2123   <- identical
```

Both runs: `+5 records`, a 40-line diff, `line_name: 106` at 2026-01…2026-05 with all three figures `0.0`; `metro_line_metadata_current.json` untouched in both. **`docs/ROADMAP.md`'s account of this gate is accurate in every particular**, including the sha prefix it quotes. Worktree restored, `git status` clean.

## The other independent checks

- `pytest scripts/` → **285 passed**. Six pre-existing test files; only `test_convert_excel_ridership.py` touched, **254 insertions, 0 deletions** — verified by counting removed lines in the diff, not by reading the body.
- `_make_bus_df_with_totals` / `_make_rail_df_with_totals` reused verbatim; neither helper appears in the diff.
- `TestDLineNotAttributedToBLine` present, three assertions, all three directions of the claim.
- Reconciliation invariant present *and* true on the real archives, not just fixtures: max absolute difference across 22 frames × 3 day types = **3.64e-12**.
- `*_ACT` dropped, `*_OFFS` kept. Docs say Boardings/Alightings for UI copy, `*_ONS`/`*_OFFS` for pipeline columns; nothing under `src/` moved, so no vocabulary leak is possible yet.
- All 98,471 emitted rows produce keys matching `^(bus|rail):[a-z0-9]+(-[a-z0-9]+)*$`, and **0** keys carry more than one display name across the whole window.

## Every factual claim in the new docs, checked against `data/raw/`

| Claim | Verdict |
|---|---|
| Union Station is `1026`/`4001`/`5001` on A/B/D in 2025-12 | exact |
| Rail leaf rows 112 → 124 at **2025-09** | exact; decomposes as 801 +4 (Foothill) and 805 0→8 (route breakout) |
| 124 → 127 at **2026-05**, D Line extension | exact; 805 8→11 |
| No existing `station_order` shifted across 2025-07…2026-05 | 0 pairs with >1 value |
| Rail carries a line-total (`ROUTE == "Total"`, `STATION_ORDER` NaN) and a route-total (`STATION_ORDER == "Total"`) | exact — inspected raw 2025-12 rail rows |
| 12 distinct suffixed rail names; `REAL_SUFFIXED_NAMES` is "every form in `data/raw/`" | exact — declared set == observed set, no over- or under-stripping, 0 names with `-` outside a platform suffix |
| 14,851 → 8,797 (2025-07…11) and 14,927 → 8,838 (2026-01…05) | exact |
| Raw-name grouping gives 8,841; normalisation folds three further pairs | exact — the three are line 60 / 611 `Pacific / RR-{-,}Xing` and line 611 `District / {,,}Produce Plaza` |
| 79% of stops serve exactly one line | 78.0–78.9% |
| Max 154 stops on one bus line | 154 through 2025-12, 153 from 2026-01 |
| 154 stops ⇒ ~77 riders of rounding | arithmetic correct |
| Alias tables ship empty because no rename occurs in the window | supported — no month has both an add and a drop; and 2026-01, the only month with drops, has **zero** additions, so nothing there could be a rename target |

---

## Findings

### F1 — `scripts/README.md`'s "bus p25 is 4.4" is a different grain (minor, docs)

The sentence sits under `aggregate_to_stop_ridership` and describes the values in *that* frame. Measured:

```
stop grain (stop x line), pooled 11 months, n=97,128:  p25 = 8.9
leaf grain (stop x line x direction),      n=164,036:  p25 = 4.8
leaf grain, 2025-07 alone:                             p25 = 4.4   <- the quoted figure
```

Per-month leaf p25 runs 4.3–5.2. So 4.4 is one month at the pre-collapse grain. The point being made survives — 92% of stop-grain `wd_ons` values are non-integral — but the number belongs to a row shape the paragraph is not about.

### F2 — `docs/ROADMAP.md` omits 2025-12, the one month that differs (minor, docs)

The payload sentence gives 2025-07…11 and 2026-01…05 and skips the month between them, which is the only outlier: **15,146 leaf → 8,953 collapsed, 6,775 keys**. 2025-12 is where 40 stops appear at once (an Altadena/Fair Oaks/Lake/Lincoln/Mariposa corridor, plus `bus:little-tokyo-arts-district-station` on line 30). A reader checking the two ranges against the archives will find a third figure and no sentence for it. Separately, "halves the payload" is a 41% reduction.

### F3 — the PR body is stale against its own diff (minor, body only)

- "`docs/README.md` does not link `docs/ROADMAP.md` … One line for whoever touches it next" — the diff **does** add that line (in `7bd4736`, correctly).
- "`pytest scripts/` → 282 passed" — actual is 285.
- "`data/raw/*.xlsx`" — `data/raw/` holds four `.zip` and no loose xlsx.

Nothing here is a code defect; it just means the body no longer describes the branch.

### F4 — the §1.3 rename guard, as specified, misses this window's one relocation-shaped event (forward-looking, PR 3)

The guard fires when a key appears **and** another disappears in the same month. The real churn straddles the boundary:

```
bus:san-vicente-fairfax       line 28, 2025-07 .. 2025-12,  wd_ons 2.8-4.2
bus:san-vicente-orange-grove  line 28, 2025-12 .. 2026-05,  wd_ons 0.9-3.7
```

Same line, same magnitude, same corridor, one month of overlap — so the add lands in 2025-12 and the drop in 2026-01 and the guard sees neither. Bus month-over-month deltas are `+40/-0` at 2025-12 and `+0/-8` at 2026-01, never both. Whether that pair is a rename or a genuine stop move, a same-month rule cannot distinguish it from silence. Worth widening to a ±1-month window, or a whole-series first/last-seen comparison, when PR 3 implements `detect_renames`. The README's "no rename occurs" claim itself still holds.

### F5 — the plan's 2% reconciliation tolerance is already breached by the real data (forward-looking, PR 3)

Rounded each stop with the `+0.5` convention, summed per line, compared against the shipped `ridership.json` for all 22 frames (1,252 line-months):

```
median 0.04%   p95 0.49%   max 2.35%
2026-01 Bus line 211:  349 vs 341   2.35%
2025-10 Bus line 211:  415 vs 406   2.22%
2025-07 Bus line 209:  381 vs 376   1.33%   (line 211 2025-07 is exactly 2.00%)
```

`scripts/README.md`'s wording — "a couple of percent" — is the accurate one; the plan's hard `< 0.02` assertion would fail on three line-months. Note also that for these eleven months `Days=1`, so the days-weighted average the README lists as the second cause contributes **nothing** here; the entire spread is per-stop rounding on small lines. Worth saying in the README, since a reader debugging PR 3's `--check` will otherwise chase the wrong cause.

### F6 — `min` on the display name picks the malformed spelling (cosmetic)

`stop_name=("stop_name", "min")` is lexicographic, so `bus:pacific-rr-xing` ships as `Pacific / RR-- Xing` rather than `Pacific / RR-Xing`. One stop, on two lines; the key is correct and stable, only the label carries the typo. The other folded key is unaffected — both `District /  Produce Plaza` variants normalise to the same display string.

### F7 — `aggregate_to_line_ridership`'s input contract widened (nit)

`extract_leaf_rows` coerces all six `LEAF_VALUE_COLS`, so the line path now requires the three `*_OFFS` columns it never reads:

```python
d = pd.DataFrame({'STOP_NAME':['A'],'LINE':[90],'DIRECTION':['IB'],
                  'WD_ONS':[1.0],'SA_ONS':[1.0],'SU_ONS':[1.0]})
base: 3 rows          new: KeyError 'WD_OFFS'
```

`_read_excel_bytes` pins the column count, so no real workbook can hit this — it only affects hand-built frames. The body calls the choice out deliberately and I agree with it; recording that the contract did move.

### F8 — the doctests in `stop_identity` never run (nit)

`normalise_stop_name` and `parse_station_order` carry `>>>` examples. There is no `pytest.ini`/`conftest.py`/`pyproject.toml` enabling `--doctest-modules`, so they are comments. Both are correct today (checked by hand); nothing keeps them that way.

---

## What I liked

`extract_leaf_rows` returning a copy and both aggregations wrapping it is the right seam — the rail total structure is genuinely two-level (`ROUTE == "Total"` with a NaN `STATION_ORDER`, plus a per-route `STATION_ORDER == "Total"`) and a second implementation would have got it wrong. Making the reconciliation invariant a test rather than a comment is what actually pins the two paths together, and it holds to 3.6e-12 on real data, not just fixtures. `_require_text` raising rather than letting `str(nan)` become `bus:nan` closes a failure mode that would have surfaced only as a wrong dot on a map. And `stop_aliases.json` shipping empty *with the reason checked against all four archives* is the right call — an alias invented for a rename that never happened is worse than no table.

*Verified with: 22-frame base-vs-head parity harness, a full re-ingest at both commits with sha256 comparison, `pytest scripts/`, and seven ad-hoc analyses over `data/raw/`. No files were modified; `src/data/` was restored after each ingest and the worktree is clean.*

mle17:
## The line 106 question is settled — the committed file is right

Neither answer in the original framing. `ridership.json` is **not stale**, and line 106 is **not missing data**. The line was discontinued, and the five zero rows a full re-ingest wants to add are wrong.

### Evidence

| | |
|---|---|
| Line 106 in the raw exports | 158 stops/month, 2025-07 … 2025-12; **absent** from 2026-01 on |
| Weekday boardings, 2025-12 | 4,062 — running normally to the last month it appears |
| Lines dropped anywhere in the 11-month window | exactly one: **106** at 2026-01 |
| Line 106 in `public/metro_lines.geojson` | **absent** (74, 105, 206 all present) |

Metro's own GTFS no longer carries it. It ended; it did not collapse.

A zero row asserts the line ran and carried nobody. Five of them would draw 106 plunging from 4,062 to a five-month flatline along the axis — which reads as a ridership collapse, not a line that stopped. `CLAUDE.md` already has the words: *a month a line doesn't report is a gap, not a zero.*

### Why a re-ingest produces them at all

`fill_missing_months` cross-joins **the batch's** month range with **the batch's** line set. Eleven narrow incremental ingests never padded a line across months outside its own batch; one full re-ingest spans all 11 months and pads everything.

Line 74 is the mirror image, and it is already in the committed file: five zeros for 2025-07 … 11, then real data from 2025-12, because its batch's range covered those months. So the file's working convention is **pad the start, not the end** — at 2026-05 there are 114 lines and not one with zero weekday ridership.

**Action for PR 3: none, beyond not "fixing" it.** Don't chase a clean `git diff src/data/ridership.json` after a full re-ingest, and don't add the five rows. The control-run gate stays as `docs/ROADMAP.md` describes it.

### One thing I found on the way, which is a real bug and is not mine to patch

`merge_ridership`'s docstring promises that "gaps in new data (within its date range) are backfilled from existing data before the concat, **preserving non-zero historical values**."

That backfill is **unreachable**. Its mask is `isnull(new) & notnull(old)`, but `fill_missing_months` has already `.fillna(0)`'d every gap, so the new side is never null — verified, 0 nulls survive it. Meanwhile `pd.concat([merged, current]).drop_duplicates(keep="first")` lets the new rows win.

So a wide-range re-ingest **can overwrite real committed ridership with zeros**, and the guard documented to prevent exactly that never runs. Today it surfaces only as five insertions, because line 106's rows happen to be absent rather than present-and-nonzero. A line that reported, paused, and resumed inside one re-ingest range would lose its real numbers silently.

It lives in `scripts/process_ridership.py`, which is out of scope for this PR. Recorded in `docs/ROADMAP.md` (`e9e932a`) so PR 3 doesn't walk into it. Happy to open an issue if that's the preferred home.

mle17:
Filed the `merge_ridership` backfill bug as #177 and cited it from `docs/ROADMAP.md`.

The issue carries one thing worth repeating here, because it makes the obvious fix wrong: **a zero is not evidence of padding.** In `01-2026-Bus.xlsx`, line 60 has 193 leaf rows all summing to `0.0` — present and genuinely reporting zero — while line 106 has 0 leaf rows and is padded by `fill_missing_months` to the same three values. Lines 202, 209, 211, 550 and 577 report `SA = 0` and `SU = 0` that month too, being weekday-only service. So a backfill mask keyed on `== 0` would silently resurrect stale figures over all of them. Reported-ness has to come from the merge indicator, never from the numbers.

Nothing in this PR changes; `process_ridership.py` stays untouched.

PR #175, opened 2026-08-15
feat(layout): put the map beside the summary and the log below both

mle17:
## What this changes

The dashboard's right-hand column used to be one tall stack: chart, summary, context log, map. The map ended up at the very bottom, under a log list that has no height limit — so you had to scroll past everything to see it.

Now:

```
┌──────────────────────────────┐
│           chart              │
├──────────────┬───────────────┤
│   summary    │      map      │   40 / 60
├──────────────┴───────────────┤
│  context logs  (scrolling)   │
└──────────────────────────────┘
```

On narrow screens (under 1024px) nothing changes shape — the four still stack in that order.

## Three things to know

**The map stays put in the JSX.** It is rendered in exactly one place, never inside an `if`. Moving it into a branch would unmount MapLibre when you deselect the last line, and the map instance has to survive that.

**The summary tiles are a 2×2 grid.** They used to sit four-across, which no longer fits — the panel is about 360px wide now. All four tiles are one size so their numbers line up.

**The context log scrolls at `32rem`.** The cap is on the list, not the panel, so the collapse button stays visible while the rows scroll under it.

## Baselines

Eleven `-linux.png` baselines regenerated (`npm run test:e2e:update:linux`). Tablet and mobile mostly unaffected — they are below the breakpoint and never had the two-column row.

Also adds ADR-0008, which records why panel layout is a plain Tailwind grid: dockview was tried, merged, and reverted in #83, and the leftover `feature/panels-*` branches still look like live work.

## Verification

`npm run lint`, `npm run test` (711 pass), `npm run build`, and the full Playwright suite (71 pass, 9 skipped) — the last run in CI's own Linux image.


mle17:
Pushed two fixes from review:

**The map now fills its pane.** It stood at a fixed 400px inside a pane stretched to the summary's height, so it floated above an arbitrary strip of empty background. `#lineMap` is `flex: 1` over a 400px floor and the pane is a flex column. Measured with seven lines selected: summary and map panes both 492px, map 428px inside its padding — previously 400px in a 492px pane. The floor still governs where there's no spare height: nothing selected, and every single-column layout below `lg`.

**The scroll cap was truncating the context-log baselines.** Caught in review — the palette shot had gone from fifteen rows to two. Its nine category assertions still passed, because Playwright counts an element scrolled out of a container as visible; it still has a box. So the image stopped covering what its docstring claims while the suite stayed green.

Both shots now lift the cap via `unclampLog`, and the cap gets an assertion of its own instead — `scrollHeight > clientHeight` with a 512px ceiling, over a window whose fifteen rows make that meaningful. Net: behaviour that was pinned only by an image is now pinned by a test, and the images are back to covering the palette.

Suite is 71 passing, up from 69 — the new assertion runs in both projects.

mle17:
Reverted the full-bleed map pane — the map keeps `.pane`'s 2rem padding again, like every other pane.

It answered the wrong problem. The complaint was that the map pane did not reach the bottom of its row, and the cause was `items-start` holding the two panes at their own heights rather than a common one. Removing that fixed it. By the time the padding came off, the pane was already full height — so dropping the inset only made the map bigger; it did not fix anything still broken. Keeping it would have left the map as the one pane in the app that frames its contents differently, for no reason that survives the diagnosis.

`#lineMap` stays elastic (`flex: 1` over a 400px floor), so it still fills whatever height the summary beside it drives — now inside the padding rather than through it.

Eleven baselines return to the padded rendering.

PR #179, opened 2026-08-15
feat(pipeline): join GTFS coordinates onto stop-level ridership

mle17:
PR 2 of 5 in the stop-level ridership batch ([roadmap](docs/ROADMAP.md)). Adds coordinates for every stop. No app changes — this PR only produces a data file for later PRs to use.

## The problem

Metro's ridership spreadsheets tell us how many people board at each stop, but not **where** the stop is. They give a name and nothing else:

```
STOP_NAME              LINE   WD_ONS
Vermont / Wilshire     204    1,203.4
```

Metro's GTFS feed (the public schedule data that already powers the route lines on our map) has the coordinates, and uses the same names. So we can match them up by name.

## What this adds

`scripts/fetch_stop_locations.py` downloads the GTFS feed, reads the stop names out of `data/raw/`, matches them, and writes `src/data/stop_locations.json`.

```bash
npm run fetch-stops
```

Both sides of the match go through `stop_identity.stop_key()` — the function PR 1 added that turns `"Vermont / Wilshire"` into `"bus:vermont-wilshire"`. Using it on both sides is what guarantees the keys here line up with the keys in the ridership data.

## What the output looks like

```json
{
  "generated_from": { "...provenance..." },
  "stops": {
    "bus:vermont-wilshire": {
      "name": "Vermont / Wilshire",
      "lat": 34.062042,
      "lon": -118.29168,
      "mode": "Bus",
      "gtfs_stop_ids": ["13000140", "14089", "5702"],
      "spread_m": 31.1
    }
  },
  "unmatched": [
    { "stop_key": "bus:allenford-sunset", "name": "Allenford / Sunset", "mode": "Bus",
      "lines": [602], "reason": "no-gtfs-match" }
  ]
}
```

A few real entries, each showing something different:

**Ordinary bus stop.** Three GTFS stops share this name — both sides of the street, plus a third nearby. They're one row in the ridership data, so they become one dot at the average position. `spread_m` is how far apart they were: 31 m, i.e. across a street.

```json
"bus:vermont-wilshire": {
  "name": "Vermont / Wilshire", "lat": 34.062042, "lon": -118.29168, "mode": "Bus",
  "gtfs_stop_ids": ["13000140", "14089", "5702"], "spread_m": 31.1
}
```

**Rail station.** GTFS models most stations as a "parent" plus one child per platform. We take the parent, so a station is one dot no matter how many platforms it has. Hence one id and `spread_m: 0`.

```json
"rail:union-station": {
  "name": "Union Station", "lat": 34.056197, "lon": -118.234249, "mode": "Rail",
  "gtfs_stop_ids": ["80214S"], "spread_m": 0.0
}
```

**A renamed station.** The spreadsheet still says `Long Beach Blvd Station`; GTFS calls it `Lynwood Station`. An alias entry connects them.

```json
"rail:lynwood-station": {
  "name": "Lynwood Station", "lat": 33.92488, "lon": -118.209945, "mode": "Rail",
  "gtfs_stop_ids": ["80312S"], "spread_m": 0.0
}
```

**A stop with no coordinates.** GTFS doesn't list it. It stays in the file under `unmatched` rather than disappearing — see below.

```json
{ "stop_key": "bus:allenford-sunset", "name": "Allenford / Sunset", "mode": "Bus",
  "lines": [602], "reason": "no-gtfs-match" }
```

## Match rate

| | Matched | Total | |
| --- | --- | --- | --- |
| Bus | 6,756 | 6,785 | 99.6% |
| Rail | 110 | 110 | 100% |

Running it twice produces a byte-identical file, so re-running it won't create a noisy diff.

Rail needed 10 alias entries in `scripts/stop_aliases.json` to reach 100%. Each was checked two ways — the spreadsheet numbers stations in order along each route, so an unknown name is pinned by its neighbours, and then the GTFS coordinates confirm it. Two would have been guessed wrong from the names alone:

- `Aviation Station` → `Aviation / Imperial Station`. GTFS has both an `Aviation / Imperial` and an `Aviation / Century`; the route order rules one out.
- `Long Beach Blvd Station` → `Lynwood Station`. No name similarity at all — the station was renamed.

Full table with the evidence is in [`scripts/README.md`](scripts/README.md).

## Two problems found along the way, both fixed

**1. The stop-grain ingest couldn't read `data/raw/` at all.**

One row in `06-2026-Bus.xlsx` (line 155, 2.9 weekday boardings) has a blank stop name. `stop_identity` refuses blank names on purpose — otherwise every nameless row on a line silently merges into one fake stop — so `aggregate_to_stop_ridership` threw on that file. Nobody hit it before because the month arrived in #176, after PR 1 was written.

Filtered in `aggregate_to_stop_ridership`, and **deliberately not in `extract_leaf_rows`**. Those riders really did board line 155; what's missing is *where*. Dropping them upstream would take them out of the line totals too and quietly change committed history in `ridership.json`. So the line keeps them, the stop grain doesn't, and the drop is printed. That's the one case where the per-line reconciliation isn't exact, and the tests say so.

**2. 39 bus stops were getting the wrong coordinates.**

Bus stops are named after intersections and LA reuses intersection names between cities. `Main / Pico` exists in downtown LA *and* Santa Monica, 21 km apart, so averaging them puts the dot in neither.

First attempt was to use each line's route shape to work out which of the two places is meant. It rescues exactly one stop (`bus:mission-broadway`, 8.5 km → 25 m, dropping stops no reporting line runs past) and that's it — because a reused name's reporting lines are the union over *both* places. `bus:main-pico` reports on lines 10, 30, 33 and 55, and those serve both intersections, so their shapes cover both. Reading `stop_times.txt` would fail for the same reason. The key is genuinely two different places, and one dot per stop key can't represent that.

So a group still more than 1 km wide gets **no coordinate**, and goes into `unmatched` with `reason: "ambiguous-name"`:

```json
{ "stop_key": "bus:main-pico", "name": "Main / Pico", "mode": "Bus",
  "lines": [10, 30, 33, 55], "reason": "ambiguous-name",
  "spread_m": 21415.6, "gtfs_stop_ids": ["13716", "3117", "3157", "5279"] }
```

That's the same "has ridership, no geometry" path the 13 renamed stops already use, so the chart and the ranked table still show it and only the map skips it. **No stop in the file now carries a coordinate known to be wrong.** The bus match rate drops from 6,772 to 6,756 as a result, which is the point — the widest remaining centroid is 899 m, down from 22.9 km.

## Choices made

- **Only stops that appear in the ridership data are written.** GTFS has 11,892 bus stops; the spreadsheets mention 6,785. Including the rest would add ~1 MB nothing reads.
- **Stops with no coordinates are kept, not dropped.** They still have ridership, so they belong in the chart and the table — they're just missing from the map. Dropping them would make a line's stop count change month to month depending on when GTFS caught up with a rename.
- **`name` is the GTFS name.** It's the current official one; the app gets its display name from the ridership rows anyway.
- **No timestamp in the file.** A timestamp would produce a diff every run even when nothing moved. A hash of the GTFS data used serves the same purpose.
- **Output stays pretty-printed** (1.6 MB). This is the file where a wrong alias silently moves a dot, so the diff needs to be readable.

## Verification

- `pytest scripts/` → **365 passed**. 61 new tests, all using fake data, no network.
- Two full runs produce a byte-identical file.
- `scripts/fetch_metro_lines.py` is untouched — its CSV helpers are imported so there's still only one CSV parser in the repo.
- Reviewed by a separate agent: no blocking findings ([comment](https://github.com/streetsforall/metro_ridership_app/pull/179#issuecomment-5303293290)); five of its six nits fixed in `852c788`. A second review of the fixes above also found no blockers ([comment](https://github.com/streetsforall/metro_ridership_app/pull/179#issuecomment-5303566982)); all eight of its nits are fixed in `a0ba6b1`.
- `scripts/convert_excel_ridership.py` is touched for fix 1 above. Line-level output is unchanged by design, so `ridership.json` is unaffected.
- `origin/main` merged in (`a2a8bc0`) — the only conflict was `docs/ROADMAP.md`, where #178 had already corrected the same coverage line.


mle17:
## Review — PR 2 geometry (plan Part 1 §1.5, §1.7 bullet 4)

**No blocking findings.** Six nits below, none of them a merge gate. Reviewed
`312c3ae..origin/feat/stop-locations`; everything below was checked against the code and
the data, not against the prose.

### Must-not-drift checklist

| # | Item | Result |
| --- | --- | --- |
| 1 | `stop_key` from `stop_identity.stop_key`, both sides | ✅ |
| 2 | Stops with ridership and no geometry retained | ✅ |
| 3 | `fetch_metro_lines.py` unmodified; CSV helpers imported | ✅ |
| 4 | Nothing outside the owned set touched | ✅ |
| 5 | Tests hit no network | ✅ |

1. GTFS side → `_collect_members` L155 calls `stop_identity.stop_key` directly. Ridership
   side → the imported `aggregate_to_stop_ridership`, which keys at
   `convert_excel_ridership.py:216`. No `_slugify`, no normaliser, no second parser
   anywhere in the new file. Verified end-to-end rather than by reading: re-running
   `collect_ridership_stops` over `data/raw/` reproduces the committed file's key set
   **exactly** — 6,895 keys = 6,882 `stops` + 13 `unmatched`, bus 6,785 / rail 110.
2. `join_locations` appends rather than skips; 13 bus stops are in the shipped file.
   Mutation — delete the `unmatched.append` — fails 4 tests.
3. Not in the diff. `fetch_gtfs` + `GTFS_URLS` imported; `parse_csv` / `split_csv_line`
   arrive transitively through `fetch_gtfs`'s closure, so there is still one CSV parser.
4. Exactly the seven owned paths. `package.json` = one line, `"fetch-stops"`.
5. `requests` is reachable only via `fetch_gtfs`, which no test calls. 39 tests, 0.9 s.

`pytest scripts/` on the branch → **324 passed**. The six pre-existing test files are
unmodified and stay green with the ten new rail aliases in place.

### The alias entries — evidence checked, not taken on trust

Re-derived the per-route `STATION_ORDER` sequences from `06-2026-Rail.xlsx` and
cross-checked each against the committed GTFS ids and coordinates. **All ten hold.**

- **`long-beach-blvd-station` → `lynwood-station`** — 803 seq 3003, between Lakewood Blvd
  (3002) and Willowbrook (3004). GTFS `80312S` @ `33.92488,-118.20995` sits between
  Lakewood Blvd `80313S` (−118.1406) and Willowbrook `80112S` (−118.2376). Confirmed.
- **`aviation-station` → `aviation-imperial-station`** — the coin flip the README calls
  out, and it lands right. 803 seq 3010 between Hawthorne / Lennox (3009) and *Aviation
  Century* (3011); Aviation / Century appears **separately** in the same sequence, so it
  is ruled out by the data, not by the name.
- **`amc-lax-station` → `lax-metro-transit-center`** — 803 seq 3012 / 807 seq 6008. GTFS
  `80702S` @ `33.95007,-118.37866` falls between Westchester / Veterans (6007) and
  Aviation / Century (6009). Confirmed.
- APU (801/1044, between Azusa Downtown and Glendora), Grand Arts (804/20), Harbor
  (803/3006), MLK (807/6002), Westchester (807/6007) — all confirmed by sequence.
- **The two reversed `expo-crenshaw-*` entries are right, and the reasoning is right.**
  The export spells both platforms `Expo / Crenshaw Station - {E,K} Line`, which
  `strip_rail_platform_suffix` already folds; GTFS spells them
  `Expo / Crenshaw {E,K}-Line Station`, which the regex cannot fold — no ` - `. So the
  export's spelling is the only clean one and folding GTFS onto it is correct.
- Checked every alias target against the export's own name set: **no alias silently
  merges two distinct export stops.**

### Determinism, and the numbers

- Committed file round-trips **byte-identically** through `json.dumps(indent=2) + "\n"`;
  `stops` keys, `unmatched`, and every `gtfs_stop_ids` list sorted; no timestamp.
- Every figure in ROADMAP / README reproduces from the file: 39 stops > 200 m, 12 > 5 km,
  4,945 multi-stop groups, median spread **41.4 m**, bus 6,772/6,785, rail 110/110.
- The `06-2026-Bus.xlsx` claim is exact — line 155, `WD_ONS` 2.9, one leaf row / two raw
  rows, and `aggregate_to_stop_ridership` really does raise
  `ValueError: Stop name is missing (nan)` on that frame. `drop_unnamed_rows` is a
  correctly-scoped local workaround and says so.
- Largest same-name group in the feed is 5 stops → the quadratic `max_pairwise_m` is fine,
  as its docstring claims.

### Test quality — mutation-tested, not eyeballed

Each of the four behaviours the test module's docstring claims was checked by breaking the
code and confirming a test goes red. Centroid → first-point: 2 fail. Bus `location_type`
filter removed: 1 fail. `parents or members` → `members`: 1 fail. `unmatched.append`
removed: 4 fail. **All four regressions are caught.** Two of them are caught by a
different test than the one named for them — see F2/F3.

---

### Findings

**F1 — `--spread-warn` never reaches the written file.** `main` passes `args.spread_warn`
to `spread_warnings`, but `build_document` hard-codes `"spread_warn_m": SPREAD_WARN_M`.
`--spread-warn 100` (a form `scripts/README.md` documents) warns at 100 and records 200.0.
Nothing reads the field yet, so the cost is a provenance value that can lie.

**F2 — the test named for the parent-station preference does not test it.** Verified by
mutation: `parents or members` → `members` leaves
`test_rail_prefers_the_parent_station_over_its_platforms` **green**. `STATION_PARENT` lat
`34.056197` and the platform centroid `34.056200` differ by 3e-6, inside `pytest.approx`'s
default 1e-6 relative tolerance (≈3.4e-5 absolute at this latitude). The behaviour is
still covered — `test_rail_records_only_the_ids_that_produced_the_coordinate` is what goes
red — so this is false confidence in one test, not a hole. Move the parent ~50 m off the
platform mean and it bites.

**F3 — `test_document_is_byte_identical_across_runs` is vacuous, and `build_document`'s
own sorts are untested.** Verified by mutation: dropping *both* `sorted(...)` calls in
`build_document` leaves all 39 green — the output stays ordered only because
`join_locations` already sorted, which is separately tested. The test compares two
identically-constructed documents in one process, so it can only catch set/hash ordering.
Handing `build_document` a deliberately unsorted dict would test what its docstring
claims (ROADMAP risk 3).

**F4 — one station ships a platform-specific display name.** `_summarise_group` takes
`min()` over the GTFS display names, so `rail:expo-crenshaw-station` carries
`"Expo / Crenshaw E-Line Station"` while the export — and the platform sign — says
`Expo / Crenshaw Station`. I diffed export-vs-GTFS names across all 6,882 matched stops:
**bus is identical on both sides everywhere**, and of the 11 rail divergences 10 are the
intended "GTFS's current name wins" cases (`APU Station` → `APU / Citrus College
Station`, etc.). This is the only one where the GTFS name is the worse of the two, and it
is the one station the alias comment calls export-canonical. PR 4/5 could prefer the
ridership-side `stop_name`, or the shortest name in the group.

**F5 — `iter_raw_frames` is quieter than the `convert_zip` it mirrors.** For a zip holding
`YYYY-MM.xlsx` inner files whose own name doesn't say Bus or Rail, `convert_zip` raises
`ValueError`; `iter_raw_frames` does `continue`. A future mis-named archive contributes
zero stops **silently** — those stops would be absent from the file altogether, not even
in `unmatched`. Low probability against the five names in `data/raw/` today, and the
docstring already flags the iteration as wanting extraction, but it is exactly the drift
the one-copy rule exists to prevent.

**F6 — doc wording.** `scripts/README.md`: *AMC / LAX Station* is confirmed by "terminus
of 803 and 807". It is the terminus of 803 (seq 3012 of 3012); on 807 it sits at 6008 of
6013, with Redondo Beach the terminus. The identity is right, the sentence isn't.

Not a finding, checked anyway: `write_text` without `newline="\n"` writes CRLF on Windows,
but that matches `compute_line_distances.py`, `fetch_metro_lines.py` and
`process_ridership.py`, and `core.autocrlf=true` normalises it. No regression.

---

Reviewed by Claude Opus 5 · no merge, no approval, no push — merge authority stays with
the repo owner.

mle17:
Review addressed in `852c788`. No blocking findings were raised; five of the six nits were cheap enough to take.

- **F1** `--spread-warn` now reaches `build_document`, so the recorded threshold is the one that was used. The committed file is unchanged — the default is 200.0 either way.
- **F2** the parent now sits ~55 m off the platform centroid instead of 3e-6 degrees, and the test asserts the fallback is *not* what was chosen. Confirmed: mutating `parents or members` → `members` now fails the test named for it.
- **F3** replaced with a test that hands `build_document` unsorted input. Confirmed: dropping either `sorted(...)` fails it. Added a companion for the F1 passthrough.
- **F5** `iter_raw_frames` now raises where `convert_zip` raises.
- **F6** AMC / LAX reworded — terminus of 803, mid-route on 807.

**F4 left alone.** `rail:expo-crenshaw-station` keeps GTFS's `Expo / Crenshaw E-Line Station` as its `name`. Shortest-name would not help (both platform spellings are the same length), and changing the rule to fix one station's label would change the documented meaning of the field for all 6,882. The app takes its display name from the ridership rows' `stop_name`, so nothing renders this — noted for PR 4/5 as the reviewer suggested.

`pytest scripts/` → **325 passed**, six pre-existing test files still unmodified.

mle17:
## Review — `a2a8bc0..bf1ce0b` (second pass, scoped to the new commit)

**Verdict: no blockers.** Every claim I was asked to verify holds, and holds against the code and the
committed data rather than against the prose. Eight nits below, none of them merge-gating.

### Verified

**1. Line-level output is unchanged — proven, not assumed.** Loaded `convert_excel_ridership` at both
`a2a8bc0` and `bf1ce0b` side by side and ran every real monthly frame in `data/raw/` through both:
24 frames, 12 Bus + 12 Rail, `assert_frame_equal(check_exact=True)` on `aggregate_to_line_ridership`
*and* on `extract_leaf_rows`. **0 mismatches.** The three hunks in that file are a new helper inserted
above `aggregate_to_line_ridership`, a docstring, and a filter inside `aggregate_to_stop_ridership`;
the line function's body and `extract_leaf_rows` are byte-identical. `ridership.json` cannot move.

The nameless row is real and it matters: `06-2026-Bus.xlsx`, line 155, `WD_ONS 2.9 / SA_ONS 4.5 /
SU_ONS 1.0`. It is the only one in the corpus. Dropping it in `extract_leaf_rows` would have restated
three day types of committed history. Filtering at stop grain instead is the right call and the
`TestUnnamedStopRows` group pins it from both directions.

**2. `stop_key` on both sides.** GTFS side `_collect_members` → `stop_identity.stop_key`; ridership
side `aggregate_to_stop_ridership` → the same. Nothing recomputes a slug locally; the only local
string work is `key.split(":", 1)[0]` for the mode prefix.

**3. Unmatched retained.** Fresh network-free re-ingest of `data/raw/` → 6,785 bus + 110 rail keys,
which equals `generated_from.stop_keys` exactly, and
`set(stops) | {u.stop_key for u in unmatched} == set(ridership keys)`. Nothing dropped, nothing
invented. The `lines` lists on every `unmatched` entry reproduce from the ingest.

**4. `fetch_metro_lines.py` unmodified** — not in the diff at all. Six symbols imported
(`GTFS_URLS`, `fetch_gtfs`, `build_shape_points`, `build_route_shapes`, `resolve_bus_route`,
`resolve_rail_route`); no copies.

**5. No network in tests.** `test_fetch_stop_locations.py` touches no fetch path. 357 tests, 3.7 s.

**6. `stop_locations.json` matches the code that generated it — checked entry by entry.**
17 stops had `spread_m > 1000` at `a2a8bc0`. Exactly one (`bus:mission-broadway`) was rescued by
`refine_by_line`: 8,518.1 m → 24.7 m, dropping ids `13731`/`5303`, landing at
`34.073673, -118.196644` — which is the real Mission Rd / N Broadway in Lincoln Heights, not a
midpoint. The other 16 lost their coordinates and moved to `unmatched` with `reason:
"ambiguous-name"`, carrying `spread_m` and `gtfs_stop_ids` forward unchanged. The 13
`no-gtfs-match` entries are byte-identical to the old `unmatched` plus the new `reason` key.
**Zero other entries changed.** Every number in `ROADMAP.md` and `scripts/README.md` reconciles:
6,756/6,785 · 110/110 · 39 above 200 m = 22 kept + 17 wide · 12 above 5 km · median 41.4 m.

### Mutation testing

Mutated the behaviour, ran the suite, reverted. 12 mutations, **11 caught**:

| Mutation | |
|---|---|
| nameless filter removed | ✅ 2 fail |
| nameless drop moved into `extract_leaf_rows` | ✅ 3 fail |
| drop made silent | ✅ 1 fail |
| `refine` returns `kept` not `kept or members` | ✅ |
| `refine` treats "no shapes" as "reject all" | ✅ |
| ambiguous-name branch removed (wrong dot written) | ✅ 2 fail |
| ambiguity judged *before* refinement | ✅ 3 fail |
| refinement not recorded in `refined` | ✅ 2 fail |
| rail parent-station preference removed | ✅ |
| bus admits every `location_type` | ✅ |
| `no-gtfs-match` stops dropped | ✅ 4 fail |
| `near_any_point` skips the haversine check | ❌ survived |

The survivor is benign: the bbox pre-filter already rejects both the 20 km case and the
tolerance case, so the exact-distance check is never the deciding vote in any test. Behaviourally
that turns the circle into a square out to `√2 × 150 m`, which for a knob documented as
"generous on purpose" changes nothing.

### On the two thresholds

`AMBIGUOUS_M = 1000` sits in a real gap, but not the one the comment describes. The comment says
*"the widest genuine one in the feed is a transit centre at ~250 m"*. The file it generated says
otherwise: **899.5 m** (`bus:sunset-marquez`), then 855.9 (`bus:montrose-florencita`), 767.1
(`bus:gayley-landfair`). The tightest ambiguous one is 1,485.4 m (`bus:foothill-osborne`). So the
gap is 899.5 ↔ 1,485.4 and 1 km sits ~100 m above the floor, not ~750 m. `scripts/README.md` gets
this right ("22 stops still centroid across 200–900 m"); the code comment is the one that is wrong,
and it is the comment a future maintainer will read before moving the constant. **Fix the comment.**

`NEAR_LINE_M = 150` only ever decided one stop, and it decided it correctly. Nothing falls on the
wrong side of it in this corpus. Fine.

### Findings

**a. The nameless-row guard screens a narrower condition than the one that raises.** `_has_stop_name`
tests None / NaN / whitespace. `stop_identity.stop_key` raises `ValueError` on anything with *no
keyable characters*, which is a strictly larger set — verified: a `STOP_NAME` of `-`, `.` or `///`
passes `_has_stop_name` and still takes `aggregate_to_stop_ridership` down with
`ValueError: Stop name '-' has no usable characters for a key.` No such row exists in `data/raw/`
today, so this is latent, not live. But the docstring and the ROADMAP now describe this class of
failure as handled, and it isn't. Note `_collect_members` in this same PR gets it right by catching
`ValueError` around `stop_key` — the two sides of the join disagree about how to ask the question.
One line: catch the exception instead of predicting it.

**b. `--spread-warn` is no longer a reporting knob.** It is passed as `threshold` into
`join_locations`, where it gates `refine_by_line`, so it now decides which stops get *geometry*.
Demonstrated: same inputs, `--spread-warn 200` → the stop is refined and mapped; `--spread-warn 50000`
→ no refinement, and the stop falls out as `ambiguous-name`. `README.md` still documents it as
`--spread-warn 100 # tighten the warning` and argparse still says "warn above this centroid spread".
Either split the two jobs, or gate refinement on `ambiguous_m` (which is the case it exists for) and
leave `--spread-warn` purely cosmetic. Second option looks right: narrowing a genuinely-wide 300 m
group is not what `refine_by_line` was built for.

**c. `refined` dies with the terminal.** The docstring's stated principle is that each narrowing is
recorded "so the run says what it changed rather than quietly moving dots". The committed file records
`refined_by_line: 1` and nothing else — no stop key, no before/after, no dropped ids. So
`bus:mission-broadway`'s entry is now indistinguishable from an ordinary two-member group, and the
evidence that its dot moved 8.5 km exists only in a console someone has since closed. The
`ambiguous-name` path does this properly (reason + `spread_m` + `gtfs_stop_ids` all committed) —
which is exactly the contrast that makes the omission stand out. One deterministic list, one entry
today. Put it in `generated_from`.

**d. `refine_by_line` earns one stop.** 17 candidates, 1 rescued, ~100 LOC plus two more GTFS tables
parsed per feed — `shapes.txt` is most of why this script downloads ~300 MB. The ROADMAP says so
plainly, which I credit; this is disclosure, not spin. It is defensible on its own terms (the
fallbacks are safe in both directions, the one rescue is correct, and it will do more work as the
alias table grows), so I am not calling for its removal — but with (c) unfixed it is complexity whose
only trace in the repo is the integer `1`.

**e. Module docstring, "fine at ~20 m, which is what 4,945 of these groups are"** — 49 groups are
≤20 m; the median is 41.4 m; and the count is now 4,929 after 16 lost their coordinates. The README's
phrasing ("4,945 stops group more than one GTFS stop and the median spread among them is 41 m") is the
accurate one.

**f. `line_shapes.update()` replaces rather than merges across feeds.** Lines 901/910 resolve in both
GTFS feeds, and the bus feed is iterated second, so the rail feed's shapes for those lines are
discarded rather than combined. Benign — `refine_by_line` falls back to every member when the filter
empties — but unguarded and non-obvious.

**g. `build_line_shapes` has no test.** It is the one new function that reaches into another module's
helpers and the one whose output feeds every refinement decision. A five-row `get_file` stub would
cover it; everything it depends on is already pure.

**h. `alias_stub(missing)` is untested.** Mutating it back to `alias_stub(unmatched)` leaves all 357
green. That regression would print a paste-ready alias stub for names GTFS *does* have — advice that
is actively wrong, since aliasing an ambiguous name cannot help.

### Judgement

The load-bearing decision in this diff — refusing to write a coordinate rather than writing a
midpoint in neither place — is the right one, and the way it was implemented is better than the
outcome required. Reusing the existing "has ridership, no geometry" path instead of inventing a
second exclusion mechanism means PR 5 has one case to handle, not two; and keeping `spread_m` and
`gtfs_stop_ids` on the excluded entries means a later reader can re-derive the decision without
re-downloading the feed. Ordering refinement before the ambiguity judgement, and pinning that order
with a named test, is the detail I would most expect a reimplementation to get wrong.

The two exclusion reasons are honestly distinguished. `no-gtfs-match` and `ambiguous-name` are
different problems with different fixes — one wants an alias, one cannot be fixed at this grain at
all — and only the first now prints an alias stub. That is the right split.

Nothing here should hold the merge. (a) is the only finding with teeth and it is latent; (b) and (c)
are small and worth doing before PR 3 builds on this file. Merge call is yours.

---
*Reviewed at `bf1ce0b`. `pytest scripts/` 357 passed. `fetch_stop_locations.py` was not executed —
no network. No files modified.*

PR #180, opened 2026-08-15
feat(stops): stop-ridership Vite plugin and the src/stops derivation

mle17:
The app currently shows ridership **per line**. This batch of PRs takes it down to **per stop**. This is PR 4 of 5 — the client-side plumbing. It ships **no visible UI**; PR 5 adds the map layer and the panel.

## Why there is no data file in this PR

The stop data files (`src/data/stop_ridership.bus.json` and `.rail.json`) are generated by **PR 3, which hasn't merged yet**.

So on this branch they don't exist — and that's fine, because the code handles it: the build serves nothing and reports "zero months of stop data". Same thing happens on a fresh clone before anyone runs the data pipeline. Tests run against small fake files in `vite/__fixtures__/`.

**Please don't read the missing file as an unfinished PR.**

## What's in it

**1. A Vite plugin — `vite/stop-ridership-plugin.ts`**

Serves the stop data files to the browser at runtime instead of baking them into the JavaScript bundle (they're several megabytes; the bus file alone is ~3.8 MB). It also exposes a tiny module telling the UI which months the data covers, so the panel can say "stop data covers Jul 2025 – May 2026" without downloading megabytes to find out.

It's a *separate* plugin from the existing `ridership-data-plugin.ts`. That one's job is rewriting the line-level data into a compact format at build time. The stop files already arrive compact from Python, so they just need serving. Combining them would put a multi-megabyte file read in front of something every single test loads.

**2. A new folder — `src/stops/`**

All the stop-level calculation lives here: decoding the data file, joining each stop to its map coordinates, filtering to the selected months, and producing the numbers and map markers the UI will draw.

It's a "sealed module" — other code imports from `src/stops/index.ts` and nothing else ([ADR-0007](docs/adr/0007-a-folder-with-an-index-is-a-sealed-module.md)). Four different parts of the UI will use this, and if each did its own filtering, the map and the table could end up disagreeing about which stops exist.

**3. One shared date filter — `isInMonthWindow`**

This is the risky bit, so it's worth explaining.

The app has an unusual rule for which months count as "inside" the date range the user picked — the last two months are excluded. It looks like a bug and isn't ([ADR-0001](docs/adr/0001-ridership-month-window-is-deliberately-offset.md)); people share URLs that depend on it and screenshot tests are pinned to it.

That rule was written inline inside the chart's code. The stop panel needs the same rule. If I'd copied it, the two panels could drift apart and show different months for the same URL. So I moved it — **unchanged, character for character** — into its own file, and both the chart and the stop code now call it. There's one copy, and `buildRidershipView.test.ts` was left untouched to prove nothing changed.

## Checks

| | |
| --- | --- |
| `npm run test` | 792 passed |
| `npm run lint` | clean |
| `npm run build` | clean |
| Screenshot tests | 80 specs, **0 baselines changed** |
| Bundle | stop data appears as a separate file, never inside the JavaScript |

The main JavaScript bundle grew by **36 bytes** — that's the one date-filter function being moved out into its own function. Nothing else changed. The lazily-loaded map chunk is byte-for-byte identical.

## Notes for whoever picks up PR 3 and PR 5

- **PR 3 must write the file format this PR reads.** It's defined in `src/@types/stops.types.ts`: `{schema: 1, cols: [...], stops: [...], rows: [...]}`, with columns named `year, month, line, stop, wd_ons, wd_offs, sa_ons, sa_offs, su_ons, su_offs`. The decoder looks columns up **by name**, so their order doesn't matter, and it throws a clear error naming any column it can't find. This format isn't written down anywhere else yet — worth putting in ADR-0008.
- **Rebased onto latest main**, which brought in PR 2's `stop_locations.json`. That file turned out to have a different shape than I'd guessed for two fields, which would have been a compile error for PR 5. Fixed, with a test that assigns the real file to the type so it can't drift again.
- **Two things I found but didn't fix**, since they're outside this PR's scope: `CLAUDE.md` says test files don't need to import `describe`/`it`/`expect`, but they do (the build fails otherwise); and the comment at `src/utils/dataDateRange.ts:17` describes the date filter as living in `App.tsx`, which hasn't been true for a while.


mle17:
## Automated review — PR #180

Scope reviewed: `git diff 312c3ae..origin/feat/stops-module`, 25 files. Read against the approved plan (Part 2 §2.1–§2.4), `CLAUDE.md` invariants, `CONTEXT.md`, `docs/ROADMAP.md`, ADR-0001/0004/0005/0007, and `vite/ridership-data-plugin.ts`.

**Verdict: approve with one fix.** The design holds up and the must-not-drift checklist passes on every item. One file in this PR is unreviewable and should be fixed before merge.

---

### Verification run

| Gate | Result |
|---|---|
| `npm run test` | 786 passed, 33 files |
| `npm run lint` | clean |
| `ANALYZE=1 npm run build` | clean |
| entry chunk `index-*.js` | 503,059 B → **503,095 B** (+36 B) |
| lazy chunk `OutputArea-*.js` | 1,080,298 B → **1,080,298 B**, byte-identical |
| `grep 'stop-ridership\|stop_key\|vermont-wilshire' dist/assets/*.js` | no match |

Base numbers from a real build of `312c3ae` extracted to a scratch tree, not inferred. +36 B on the entry chunk is `isInMonthWindow` becoming its own minified function — `buildRidershipView` was already in that chunk. No payload leak.

---

### Must-not-drift checklist

1. **Month Window — one copy, unchanged.** ✅ `monthWindow.ts:37-43` is `312c3ae:buildRidershipView.ts:107-112` character for character; only `continue` → `return false`. `grep 'new Date(.*\.year, .*\.month)'` over non-test `src/` → one production hit, `monthWindow.ts:37`. Callers: `buildRidershipView.ts:104`, `buildStopView.ts:144,152`. Barrel re-export carries a doc comment that says *why* it is exported, which is the right guard.
2. **`src/stops/` sealed.** ✅ Nothing outside the folder imports its internals. Nothing outside imports it at all yet — no UI. `index.ts` names the whole surface.
3. **Columns by name.** ✅ Both decoders — `stopData.ts:93-99` and the plugin's `readPayload` at `stop-ridership-plugin.ts:128-133`. `stopData.test.ts` states the fixture's `cols` in scrambled order, so a positional decoder fails the suite instead of shipping alightings labelled as boardings. That is the right way to pin it.
4. **`DayOfWeek` not widened.** ✅ `metrics.types.ts` untouched. Translation is the private `DAY_COLUMNS` record in `stopMetrics.ts:12-19`.
5. **Absent data files tolerated.** ✅ `ENOENT` → `null`, everything else rethrown. Manifest reports `{null, null, 0, 0, 0}`. Build passes on this branch, where neither file exists. Malformed still throws — absent is a state, corrupt is a bug.
6. **No inlining, lazy-load intact.** ✅ See the table. `OutputArea` byte-identical.
7. **No baseline moved.** ✅ No `.png`, no `e2e/` in the diff.
8. **No scope violation.** ✅ Changed-file list is exactly the allowed set. Nothing from §2.5–§2.7 present — no `Map.tsx`, no `OutputArea`, no `useUserDashboardInput`, no `queryParams`, no `mapPopup`. The marker GeoJSON and the sqrt radius scale are §2.5 *text* but §2.3 *files*, and §2.5 explicitly delegates the scale to `buildStopView` "where the domain is known" — correct side of the line.

---

### Blocking

**1. `src/stops/stopReadouts.ts` contains two literal NUL bytes and git treats the file as binary.**

Byte offsets 2485 and 2614 — inside the doc comment and inside the template literal:

```ts
/** Group key. `<U+0000>` cannot occur in a line id or a slug, so it cannot collide. */
const groupKey = (lineId: number, stopKey: string) => `${lineId}<U+0000>${stopKey}`;
```

Consequences, confirmed rather than assumed:

- `git diff --stat` → `src/stops/stopReadouts.ts | Bin 0 -> 5586 bytes`
- `gh pr diff 180` → `Binary files /dev/null and b/src/stops/stopReadouts.ts differ`

All 147 lines of the file are invisible on the PR page and in every future `git diff` or `git blame -L`, and the file will never 3-way merge. In a repo where the review contract is the diff, a source file git refuses to diff is a defect in its own right — and this is the file carrying `shareOfLine`, the readout ordering and the ADR-0005 parallel, i.e. the part most worth reading.

Separator choice is fine. Write it escaped:

```ts
/** Group key. `` cannot occur in a line id or a slug, so it cannot collide. */
const groupKey = (lineId: number, stopKey: string) => `${lineId}${stopKey}`;
```

Identical at runtime, and the file becomes text. (I reviewed the content out of band — it is correct. `rankA !== rankB` guards the `Infinity - Infinity → NaN` comparator case; `...metrics` on `null` writes no keys and collides with nothing spread before it.)

---

### Observations — not blocking

**2. The manifest is not cheap on the server side.** `load()` → `manifest()` → `get('bus')` reads *and* `JSON.parse`s the full payload, then walks every row to build the month `Set`. Free today because the files are absent. After PR 3 that is a ~3.8 MB read, parse and ~97k-row scan per Vite and per Vitest process, incurred by the very module whose stated purpose is to answer the coverage question *without* paying for the payload. The header's actual claim — that this stays off `virtual:ridership-bounds`' path — still holds, so nothing here is wrong. But rows are sorted by `["year","month",…]` (plan §1.6), so first-row/last-row plus a `monthCount` field written by Python would make it genuinely cheap. Worth deciding before PR 3 freezes the writer.

**3. The vitest registration is asserted but not exercised.** `vitest.config.ts`'s comment says without the plugin "every test that reaches the stop panel dies at import time." True, and untested — no spec imports `virtual:stop-ridership-manifest`, so deleting `stopRidershipPlugin()` from `vitest.config.ts` today leaves 786 tests green. One line (`import { minMonth } from 'virtual:stop-ridership-manifest'` in a spec) pins the parallel-registration rule the same way the ridership plugin's is pinned through `dataDateRange.ts`.

**4. `mode` has two contradictory contracts.** `stops.types.ts:32-38` says `mode` is the key prefix, "which magnitude domain the marker radius is normalised against." `buildStopView.ts:109-112` promises G/J Line BRT stops "therefore sit in the bus domain." But `attachStopLocations` lets `stop_locations.json` override it (`stopData.ts:170`), and `stopData.test.ts:189-194` pins that override. If PR 2's `fetch_stop_locations.py` derives `mode` from GTFS route type rather than from the stop key, every BRT stop flips to `Rail` and normalises against the rail domain — the exact outcome `buildStopView`'s comment says cannot happen. `fetch_stop_locations.py` is not written yet, so this is the cheap moment: either drop the override, or amend the two doc comments to say the location file is authoritative and the BRT guarantee is conditional on it.

**5. The wire contract lives only in this PR.** `schema: 1`; columns named `year`, `month`, `line`, `stop`, `wd_ons`…`su_offs`; `stops[]` entries `{key, name, station_order?}`. `docs/ROADMAP.md`'s "The contract PR 1 freezes" states the *pandas frame* columns (`stop_key`, `stop_name`, `mode`) — not the wire — so nothing outside `src/stops/` and four fixture files records what `write_stop_ridership` must emit. PR 3 misses a name → `decodeStopRidership` throws at runtime with a good message, but the mismatch is only discoverable by running it. Docs are outside this PR's permitted file list so this is correctly deferred; flagging it as the highest-value item for the docs PR.

**6. `monthsReported` counts records, not months with figures.** `records.length`, so an all-null row counts as a reported month and drags the average down — `stopMetrics.test.ts:31-44` pins exactly that. Faithful to `lineMetrics` (checked line by line: same `?? 0`, same divisor) and stated plainly in the function's doc. The *field's* doc — "How many months this stop reports for this line" — reads as the other thing. A table cell rendering "11 months" for a stop with 4 real months would be wrong, so worth tightening the wording before §2.6 puts it on screen.

**7. Row arity is unvalidated.** `decodeStopRidership` throws by name on a missing column and on an out-of-range stop index, but a short row yields `undefined` in a `number | null` slot, which `stopMetrics` then absorbs as `?? 0`. Silent zeroes, from the module whose own comment two functions earlier argues for "loud rather than lenient." One `row.length !== cols.length` check closes it.

**8. Nothing pins `isInMonthWindow` ≡ `containsOffset`.** They *are* equivalent — with bounds `new Date(y, m-1)` and records `new Date(r.year, r.month)`, `S < R+1 < E` ⇔ `S ≤ R ≤ E−2`. `monthWindow.ts` names `containsOffset` as the landing site for the #144/#145/#146 migration, and ADR-0001 says the boundary tests, not the arithmetic, are what makes that rewrite safe. Both specs currently state the boundaries independently and never compare the two functions, so the migration would still be a leap. A short property test over a few hundred `(window, record)` pairs asserting they agree would close it for good. Related nit: `monthWindow.test.ts`'s header says `month.test.ts` "asserts the same boundaries" — true of the rule, not of the cases; the cross-year blocks use different windows (2019-11→2020-02 vs 2019-11→2020-03).

---

### What is good

- The Month Window extraction is done the way ADR-0001 asks: verbatim copy, one production copy, both derivations importing it, and the reason recorded at the export site rather than only at the definition.
- Doc comments consistently say *why*, and every load-bearing claim I spot-checked is true — the `fileURLToPath`-not-`new URL` note, the ROUTE/BRT split, the `?? 0` divisor lifted from `lineMetrics`, the `#144/#145/#146` reference.
- Test fixtures are small, committed, and deliberately hostile — scrambled `cols`, a null-filled row, a dangling stop index, a rail-only directory, a directory that does not exist.
- Absent-data tolerance is treated as a supported state with its own describe block, not an accident.
- Interfaces are honest about their nullability: `records: null` is loading, `lat: null` is unmatched-but-kept, `stopMetrics` returns `null` for empty, `shareOfLine` is `undefined` at a zero total.

---

*Automated review. No files were modified. Merge authority remains with the repository owner.*

mle17:
Thanks — acted on all of it. Branch force-pushed: `7edea6d` → `20d6044`, one clean commit rather than a follow-up, because a second commit would have fixed the PR page while leaving the binary blob in history, so `git blame -L` on that file would have stayed broken across it.

**Blocking (1) — fixed.** The two NUL bytes are gone; they came from a literal `\u0000` surviving into the file rather than the escape. Now `\u0000` in both the comment and the template literal, identical at runtime. `git diff --stat 312c3ae..origin/feat/stops-module -- src/stops/stopReadouts.ts` → `148 ++++`, and the PR diff has no binary files at all.

**(4) mode contract — fixed, by dropping the override rather than amending the comment.** You were right that the two contracts contradicted each other, and the override was the wrong half. `attachStopLocations` now takes coordinates only; `StopLocationEntry.mode` is deliberately unread. The key prefix answers "which export did this row arrive in" by construction, and that is the whole meaning of the field. The test that pinned the override now pins the opposite, with the BRT reasoning in it.

**(7) row arity — fixed.** `row.length !== cols.length` throws and names both counts, with a test. Your reading was right: the silent path went `undefined` → `?? 0` → a column of zeroes on the map, from the module that argues for loud-not-lenient two functions earlier.

**(3) vitest registration — fixed.** `stopData.test.ts` now imports `monthCount` from `virtual:stop-ridership-manifest` and asserts the documented zero-coverage value. Deleting `stopRidershipPlugin()` from `vitest.config.ts` now fails the suite at import time instead of leaving it green.

**(8) `isInMonthWindow` ≡ `containsOffset` — closed, exhaustively.** Not a property test in the end — the space is small enough to enumerate. Every month 2018–2027 as record, against every window with endpoints in the same span: 120³ = 1,728,000 comparisons, both functions, asserting agreement. That is the thing ADR-0001 says has to hold when #144/#145/#146 replaces the arithmetic, and it now fails before anything reaches a chart baseline. Also fixed the header's overclaim about `month.test.ts` — it states the same *rule*, not the same cases.

**(6) `monthsReported` — reworded, not rebehaved.** The behaviour is faithful to `lineMetrics` on purpose and stays. The field doc now says it counts *records*, is the averages' divisor, and is not "months with a figure" — with the "11 months for a stop with 4" caption you described as the thing not to write.

**(2) manifest cost — agreed, and it is a PR 3 decision, so flagging rather than pre-empting.** Free today because the files are absent; after PR 3 it is a full read, parse and ~97k-row scan per Vite and per Vitest process, from the module whose stated purpose is answering the coverage question without paying for the payload. Rows are sorted by `["year","month",…]` per plan §1.6, so first-row/last-row plus a `monthCount` written by Python makes it genuinely cheap. I have not built that in, because relying on the sort before the writer exists would be trusting a contract nobody has written yet — if PR 3 emitted unsorted rows the manifest would silently report a wrong window, which is worse than slow. Left for PR 3 to decide with me.

**(5) wire contract lives only here — agreed, and it is the highest-value item for the docs PR.** `docs/ROADMAP.md`'s frozen contract is the pandas frame, not the wire. Docs are outside this PR's permitted file list. It is already in the PR body under "Decisions I made where the plan left room" as the PR 3 coordination point; it wants an ADR-0008 section too.

Gates re-run on `20d6044`: `npm run test` 789 passed / 33 files, `npm run lint` clean, `npm run build` clean, `ANALYZE=1` unchanged (entry 503,095 / `OutputArea` 1,080,298, no stops symbol in any chunk), and the full Playwright suite re-run in `mcr.microsoft.com/playwright:v1.62.1-noble` — 69 passed, 9 skipped, no `.png` touched.

mle17:
Round 2 taken. Two of the observations were cheap and are now in — `20d6044` → `3640ba2`, comment-only, amended rather than stacked so the branch stays one commit.

**A — dangling `ADR-0008` citation: fixed.** You were right that it pointed at a file that does not exist. Rather than drop the parenthetical and lose the reasoning, the plugin header now states the argument inline and cites `docs/ROADMAP.md`'s "Decisions that are settled", which does exist and does say it. When the docs PR lands 0008 it can take the citation back.

**B — `StopReadout` wider than any reachable state: documented, not narrowed.** Your diagnosis is right: a group exists only because a record landed in it, so `stopMetrics` never returns `null` on that path and the `Partial<StopMetrics>` half is unreachable through `buildStopView`. I have left the type alone and said so at `measuredAverage`, because the nullability is ADR-0004's contract rather than a case to render — and the note now says explicitly that a "no figures" branch in the §2.6 table would be dead code no test could reach. Narrowing the type would make `stopMetrics` and `buildStopReadouts` disagree about their own contract for the sake of a state neither can produce. This is the same posture `buildRidershipView` takes toward an empty series, which ADR-0004 calls "unreachable in production" and declares anyway.

**C — manifest field set declared twice: accepted, not closed.** Genuinely two independent lists with no compile-time link. I looked at closing it and did not like either route: `src/@types/virtual-modules.d.ts` importing `StopManifest` would make the file a module and stop `declare module` being ambient, and it crosses `tsconfig.app` → `tsconfig.node`. Building the exports from `Object.entries` moves the failure but does not remove it. As you say, `ridership-data-plugin` has the identical shape, so closing it here would make the two plugins diverge for no gain. Worth doing once, for both, when someone owns that file.

**D, E, F, G** — agreed as stated. E in particular: the wire shape *is* a transformation of the frozen frame (`mode` dropped as derivable from the key prefix; `stop_key`/`stop_name`/`station_order` dictionary-encoded behind an integer `stop`), and recording that is the highest-value line in the docs PR. F and G are outside this PR's permitted file list and are flagged in the PR body.

Gates on `3640ba2`: `npm run test` 789 passed / 33 files, `npm run lint` clean, `npm run build` clean, `ANALYZE=1` unchanged — entry `503,095`, `OutputArea` `1,080,298`, no stops symbol in any chunk. Playwright was not re-run for this amend and I am not claiming it was: the only delta from the run at `20d6044` (69 passed, 9 skipped, no `.png` touched) is twelve lines of comment across two files, which cannot reach the DOM.

mle17:
Rebased onto `origin/main` `226a2e2` — `3640ba2` → `e867cf7`, clean, no conflicts. Main moved three commits while this was open, and one of them found a real bug in here.

**#179 landed `src/data/stop_locations.json`, and my `StopLocationsFile` was wrong about two fields.** It was declared before that script existed, so it was a guess. The guess held for `stops` — `{name, lat, lon, mode, gtfs_stop_ids, spread_m}` is exactly right, 6,866 entries, none missing a coordinate — and was wrong for the other two:

- `generated_from` is not a string. It is an object carrying the GTFS feeds, the ridership archives, match rates and thresholds. Now `unknown`: nothing in the app reads it, and pinning its shape would mean editing this file every time the script records one more diagnostic.
- `unmatched` is not `string[]`. It is `{stop_key, name, mode, lines, reason}[]`, 29 entries, reasons `no-gtfs-match` and `ambiguous-name`. Now a named `UnmatchedStop`.

**That was a compile break waiting for PR 5, not a cosmetic mismatch.** `import stopLocations from '../data/stop_locations.json'` was not assignable to `StopLocationsFile` — `tsc` rejected it on `generated_from`, and then again on `mode`, because TypeScript widens a JSON file's string literals so `string` does not satisfy `'Bus' | 'Rail'`. `StopLocationEntry.mode` and `UnmatchedStop.mode` are therefore `string` now, with the reason at the declaration. No loss: `attachStopLocations` deliberately does not read that field, and `StopPlace` — the app-side shape, where the decoder controls the value — keeps the union.

There is now a guard so this cannot drift again: `stopData.test.ts` assigns the committed file to the type (the assignment *is* the assertion — it is a compile error if the shape moves) and checks every located stop has a coordinate and every unmatched entry names its lines. This is the one place the suite reads real data and it earns it, because the file is written by a Python script in another PR and nothing else would notice them diverging. Safe to import in a way `stop_ridership.bus.json` would not be: bundled metadata, not the payload fetched at runtime.

**A side benefit — the round-1 decision to ignore the location file's `mode` is now empirically vindicated rather than merely argued.** Across all 6,866 entries it never disagrees with the key prefix (0 disagreements), so dropping the override changed no behaviour; and the test above pins that, so if a future GTFS refresh ever does disagree, ignoring it is what stops a G/J Line BRT stop being normalised against the rail radius domain.

Re-ran everything on the new base, since #175's layout change moved main's own baselines and my earlier evidence was against the old set:

| Gate | Result |
| --- | --- |
| `npm run test` | 792 passed, 33 files |
| `npm run lint` | clean |
| `npm run build` (`tsc -b`) | clean |
| Playwright | **80 specs, 71 passed, 9 skipped, no `.png` touched** |

Bundle, against a fresh control build of `226a2e2`:

| Chunk | main `226a2e2` | this branch |
| --- | --- | --- |
| `assets/index-*.js` (entry) | 503,059 | 503,095 (**+36 B**) |
| `assets/OutputArea-*.js` | 1,080,479 | 1,080,479 |

Same +36 B as before — the extracted `isInMonthWindow`, unchanged by the rebase. No stop payload, no stops symbol, and no `stop_locations.json` content in any `dist/assets/*.js`.

PR #190, opened 2026-08-15
feat(stops): merge stop-level ridership and commit the two payloads

mle17:
## What this PR does

Adds stop-level ridership data to the repo.

The app shows how many people ride each **line**. The Excel files Metro sends us are
actually broken down by **stop** — we were adding those up and throwing the detail away.
An earlier PR added the code that reads stops out of those files. This PR adds the step
that saves them, and commits the result: two new data files.

```
src/data/stop_ridership.bus.json     105,984 rows   6,785 stops   109 lines   5.3 MB
src/data/stop_ridership.rail.json      1,470 rows     110 stops     6 lines    89 KB
```

Nothing renders them yet — that's the next PR. This one just produces the data.

**Existing line data is untouched.** `src/data/ridership.json` is byte-for-byte identical
before and after.

## Coverage

**July 2025 – June 2026, twelve months.** That is all the stop-level data that exists;
everything older was only ever delivered per line. Both boardings and alightings, for
weekday / Saturday / Sunday.

## What you run

Nothing changes for normal use — `python scripts/update_ridership.py` now updates both the
line data and the stop data from the same scan of `data/raw/`.

| | |
| --- | --- |
| `scripts/stop_ridership.py` | new — reads, writes and merges the two data files |
| `update_ridership.py --no-stops` | skip the new step |
| `stop_ridership.py --check` | compare stop totals against line totals |

## The file format

Column names in a header, rows as plain arrays, stop names in a separate lookup list:

```json
{"schema": 1,
"cols": ["year", "month", "line", "stop", "wd_ons", "wd_offs", "sa_ons", "sa_offs", "su_ons", "su_offs"],
"stops": [
{"key": "rail:103rd-street-watts-towers-station", "name": "103rd Street / Watts Towers Station", "station_order": 1013},
...
"rows": [
[2025, 7, 801, 0, 1699, 1747, 1728, 1918, 1001, 962],
```

- `stop` is a position in the `stops` list above — `0` is the first entry.
- `ons` = boardings, `offs` = alightings. `wd`/`sa`/`su` = weekday/Saturday/Sunday.
- It's a spreadsheet rather than a list of objects because at 106,000 rows, repeating
  `"stop_key":` on every row would be most of the file.
- One row per line of text, so a diff shows which months and stops actually changed
  instead of "the whole file".

`src/stops/stopData.ts` (already on main) is the reader. It was written first, so the
format here had to match it exactly.

## Two things a reviewer should know

**Stop totals don't exactly match line totals, and that's expected.**

Metro rounds by adding 0.5 and truncating. We round each stop; the line total gets rounded
once. Rounding 70 small numbers and adding them up isn't the same as adding them up and
rounding once.

```
3,978 comparisons    median 0.06%    p95 0.83%    max 5.10%
```

Worst case is line 602's Sunday total for August 2025 — stops sum to 103, the line says 98.
Small lines are worst, because a few riders is a big share of a small total. Forcing exact
agreement would mean either rounding stops to match a total they didn't produce, or
re-deriving 17 years of line history from stop data. `--check` reports the gap instead.

**If Metro renames a stop, the ingest stops and tells you.**

A rename looks like one stop vanishing and another appearing. Left alone that splits one
station's history into two half-height series, and nothing warns you. So when a stop
appears in the same month another disappears, the run fails and prints both lists. Fix it
by adding the rename to `scripts/stop_aliases.json`, or pass `--allow-new-stops` if they
really are new stops.

It doesn't fire on any of these twelve months.

## How the data was checked

- **It decodes.** Ran the real client reader over both committed files: 105,984 + 1,470
  records, schema accepted.
- **Stop names match the map data.** Joining against `src/data/stop_locations.json` gives
  6,756/6,785 bus and 110/110 rail — the same match rate that file reported when it was
  built, so both sides agree on what each stop is called.
- **Two full runs produce identical files**, including after deleting both and
  regenerating from scratch.
- **No double-counting** when a month is present as both a loose `.xlsx` and inside a zip.
- **One row is dropped:** `06-2026-Bus.xlsx` has a line 155 row with 2.9 weekday boardings
  and a blank stop name. It still counts toward the line total — we know those riders
  boarded, just not where.

```
pytest scripts/     409 passed        npm run lint     clean
npx vitest run      864 passed        npm run build    clean
```

## One file changed outside the pipeline

`src/stops/__tests__/stopData.test.ts` asserted `monthCount === 0`, with a comment saying
the data files "are not on this branch". They are now, so it's `12` and the test failed by
construction. It now asserts a non-negative integer, so it won't break again next month.

## Noted, not fixed

- `iter_raw_frames` (the loop over `data/raw/`) lives in `fetch_stop_locations.py`, whose
  own comment says it belongs in `convert_excel_ridership.py`. Imported rather than copied.
- `vite/stop-ridership-plugin.ts` has a comment saying these files are minified. They're
  not any more. Behaviour is unaffected.


mle17:
## Review — approve

Data-provenance review against `574b70e` (`origin/main`). **No blocking findings.** Every
claim in the PR body that I checked reproduced exactly, including the ones that would have
been easiest to round in the author's favour.

---

### The must-not-drift checklist

**1. The payload validates against `STOP_WIRE_SCHEMA = 1` and `stopData.ts` decodes it.**
Verified by decoding, not by reading the writer. A throwaway vitest spec read both
committed files off disk and ran `decodeStopRidership` → `attachStopLocations`, then
deleted (`git status` clean afterwards).

| | schema | records | places | lines | months | geometry |
| --- | --- | --- | --- | --- | --- | --- |
| bus | 1 | 105,984 | 6,785 | 109 | 12 (2025-07 → 2026-06) | 6,756/6,785 |
| rail | 1 | 1,470 | 110 | 6 | 12 (2025-07 → 2026-06) | 110/110 |

Every value is a non-negative integer; no nulls. Every row is exactly 10 wide, so the
decoder's short-row throw never fires. `place.mode === modeFromStopKey(place.key)` for all
6,895 places. The two dictionaries are disjoint. The geometry rate is PR 2's number to the
stop.

**2. `stop` is an index; `mode` is absent; `station_order` is not an identity.**
`cols` is `["year","month","line","stop","wd_ons","wd_offs","sa_ons","sa_offs","su_ons","su_offs"]`
in both files — no `mode`, no `stop_key`, no `station_order`, no `lat`/`lon`. Stop indices
run 0…6784 and 0…109, every dictionary entry is referenced by at least one row (no
orphans), and the decoder throws on an index it can't resolve. Stop entries have the
uniform key shape `('key','name','station_order')` across all 6,895 of them — no
coordinates in the payloads, which is right: the join is client-side in
`attachStopLocations`, so the plan's `attach_locations` correctly does not exist.
`grep` over `src/`, `scripts/` and `vite/` finds `station_order`/`stationOrder` only as a
carried attribute (`stopData.ts:107`) — never as a key, a lookup or a join column.

**3. Rows and the dictionary are both deterministically sorted.** The probe asserted rows
non-decreasing under `[year, month, line, stop_key]` and `keys === [...keys].sort()`, both
files. Stronger check: `load_stop_ridership` → `write_stop_ridership` over the committed
files reproduces them **byte-identically** — 5,340,669 B and 89,366 B, matching
`git cat-file -s` on both blobs. So the write path is idempotent and the sort is stable
over what shipped.

**4. Dedupe on `STOP_KEYS`.** Present twice: `load_new_stop_ridership` (`stop_ridership.py:481`)
dedupes the concat of all monthly frames before the merge, and `merge_stop_ridership`
(`stop_ridership.py:272`) dedupes the concat with the committed rows.
`test_loose_xlsx_and_its_zip_do_not_double_count` (`test_stop_ridership.py:386`) builds the
exact loose-plus-zip overlap and asserts one row survives with the right value.
Empirically: **0 duplicate `[year,month,line,stop_key]` tuples across all 107,454 rows.**

**5. `src/data/ridership.json` is untouched.** Absent from `git diff --name-status`, and the
blob sha is identical on both sides: `b442917a62154f80385c6c9562977375babf81c9`.
`metro_line_metadata_current.json` and `stop_locations.json` likewise.

**6. Files touched.** Ten. Nine are on the agreed list. The tenth is
`src/stops/stopData.test.ts`, +7/−4, one assertion. **Justified.** `monthCount` comes from
`virtual:stop-ridership-manifest`, which derives it from the payloads, so landing them made
`expect(monthCount).toBe(0)` fail by construction — the alternative was red CI. The
replacement (integer ≥ 0) is also better than re-pinning to `12`, which would fail again on
the next ingest. Disclosed in both the PR body and ROADMAP.

---

### Reproducing the claims

| Claim | Result |
| --- | --- |
| `stop_ridership.py --check` | reproduced **verbatim** — 3,978 comparisons over 107,454 stop rows, median 0.06% / p95 0.83% / max 5.10%, 26 at or above 2%, same top three |
| `pytest scripts/` | 409 passed |
| `npx vitest run` | 807 passed, 34 files |
| `npm run lint` | clean |
| `npm run build` | clean; both payloads emitted as assets (bus gzip 1,182.96 kB); `grep` for stop keys in `dist/assets/*.js` → 0 hits in both chunks |
| rename guard silent on twelve months | confirmed by running `detect_renames` over the committed payloads: bus and rail both **0** at `window=0`; `window=1` fires exactly once, `2025-12: 40 appeared / 8 disappeared` — the advisory line quoted in the PR body |

The one-sided-drift argument holds up under its own evidence: **all 26 exceedances have
stops > line, none the other way.** That directional consistency is a better argument that
this is half-up rounding rather than a forked leaf-row rule than any of the magnitudes are.

Plan §1.7 asks five things of `test_stop_ridership.py` and all five exist and test what
they claim: round-trip (`:86`), append-only byte-identity (`:231` — it asserts the committed
row *text* appears verbatim in the rewritten file, not merely that the values match),
`--overwrite` (`:256`, `:423`), alias collapse (`:339`), `detect_renames` fires and is silent
in month 1 (`:286`, `:295`). Four more that §1.7 didn't ask for earn their place: skip-a-month
(`:310`), drop-with-no-add (`:305`), rounding idempotence (`:187`), determinism under input
reordering (`:148`).

---

### Non-blocking

**a. Line endings — the one thing I'd fix soon.** `write_stop_ridership` uses
`Path.write_text`, which translates `\n` → `\r\n` on Windows. The committed blobs are LF,
so `core.autocrlf=true` covered for it here — but the repo has **no `.gitattributes`**, so a
Windows contributor with `autocrlf=false` commits a 112,775-line whitespace-only diff and
the reviewable-diff property this file's format exists to buy is gone in one push. One-line
fix, either end: `newline="\n"` on the write, or `*.json text eol=lf`. Same cause as
`npm run build` reporting the asset at 5,453 kB rather than 5,341 kB on Windows.

**b. Nine rail stations will display their pre-rename name.** The dictionary keeps `min()`
of the observed names while `stop_key` is the alias-canonicalised slug, so key and label
disagree in every aliased case: `rail:apu-citrus-college-station` → `"APU Station"`,
`rail:lynwood-station` → `"Long Beach Blvd Station"`, `rail:lax-metro-transit-center` →
`"AMC / LAX Station"`, and six more. Harmless here — nothing renders yet — but PR 5 puts
those labels on the map and in the ranked table. Worth deciding before then whether the
display name should come from the most recent month, or from the canonical side of
`stop_aliases.json`, rather than from a lexicographic tiebreak.

**c. `--check` exits 1 at its default tolerance.** The reasoning for keeping `0.02` is
sound, but it means the headline gate can't be wired into CI as-is, and `--tolerance 0.06`
lives only in prose. A named-exceptions list would let a *new* exceedance fail while the 26
known ones pass — which is the behaviour the argument in `scripts/README.md` actually wants.

**d. `_round_metro` is `floor(x+0.5)`; `compute_ridership` is `int(x+0.5)`.** Equivalent on
non-negatives, and the probe confirmed zero negative values in either payload, so this is
fine. Worth one clause in the docstring naming the assumption, since the whole
reconciliation argument rests on the two rules being the same rule.

**e. `from fetch_stop_locations import iter_raw_frames`** and the stale
`vite/stop-ridership-plugin.ts:87` comment — both flagged in the PR body rather than
patched. Agree with both dispositions for a data-only PR; importing beats copying the
archive-layout rule.

**f.** `test_update_ridership.py`'s autouse `stop_paths` fixture is the right guard, and it
works: `git status` was clean after both `pytest scripts/` and the full vitest run.

PR #223, opened 2026-08-16
feat(stops): the stop panel and the map's circle layer

mle17:
## What this adds

The dashboard has only ever measured the line. This PR adds the stop: a ranked table of individual stops, a chart of the ones you pick out of it, and circles on the map marking those same stops. It is PR 5 of 5, and the last of the batch.

Four earlier PRs (#173, #179, #180, #190) built the stop-grain data and shipped it weeks ago, and nothing has displayed it. `src/stops/` already computes everything shown here, so this PR calls `buildStopView` once and renders what comes back — it recomputes no ridership, no circle sizes, no colours and no date ranges.

Tick **Stop Ridership** under *Panel Visibility*, or put `?stops=1` in the URL. Those are one piece of state, so either route updates the other.

Ranked stop table

The table is the primary readout, and it never needs the map. It also works the way the line selector does, because the two had drifted apart: a checkbox per row, a search bar above them, `Select All` and `Clear All` beneath. Those are `LineFilters`'s three controls in `LineFilters`'s arrangement, and they carry its asymmetry too — `Select All` reaches only the rows the search currently lists and adds to what you have already picked, while `Clear All` clears everything and leaves your search text alone.

## Where the new code sits

The accented boxes are this PR; the muted ones already existed and are drawn to show what the stop grain runs beside. Every box names a symbol and the file it lives in, so a box is also a place to start reading.

01-where-the-new-code-sits

`src/stops/` is a sealed module ([ADR-0007](https://github.com/streetsforall/metro_ridership_app/blob/feat/stop-panel/docs/adr/0007-a-folder-with-an-index-is-a-sealed-module.md)), so everything above reaches it through `index.ts` and nothing imports `buildStopView` or `stopData` directly. The seam is there because four consumers — the table, the figure, the sparkline and the map — must agree about which stops exist and which months are on screen.

## Picking stops, and comparing their histories

Ticking a row, or clicking a circle on the map, draws that stop's ridership over time. Pick several and they draw together, so you can compare a corridor rather than read one stop at a time.

Two stops selected, drawn together

Every route in is a toggle: the row, its checkbox and its map circle each add a stop or remove it, and `Clear All` empties the selection outright. Those three surfaces cannot drift apart, because there is one selection and no copy of it. The URL is written from that same selection rather than read back into it, so a shared link reproduces the view without a second source of truth.

02-picking-stops

## Circles on the map

One circle per stop, drawn above the route lines. Size says how busy the stop is and colour says which line it belongs to, so the measure is left to the fill: boardings draw as a solid disc, alightings as a hollow ring, and both as both.

Stop circles on the map

## Colour has to mean two things, and that is deliberate

Everywhere else in this dashboard, colour means which line. In the chart above the stop table it means **which stop**, and that inconsistency is the one decision here a reader should expect an argument for.

The encoding ran out of channels. Hue belonged to the line and dash to the measure, so two stops on line 204 — the common case, since the table exists to rank one line's stops — drew as two identical teal lines. Each picked stop now takes a hue from an eight-colour palette, and the legend names the stop and its line beside it.

The palette stops at that chart. The row sparkline stays in its line's colour, and every selected circle keeps the neutral ring it always had. Colouring the rings would turn the map into a legend for the chart, which is a real gain, but the circles are already saturated line colours at radii from 4 to 22 pixels, so a second meaning in the ring would compete with them rather than clarify. `Map.test.tsx` asserts the ring is one colour, so extending the palette onto the map fails a test rather than slipping in unread. [ADR-0014](https://github.com/streetsforall/metro_ridership_app/blob/feat/stop-panel/docs/adr/0014-colour-in-the-stop-series-chart-means-which-stop.md) carries the trade-off and the two alternatives it beat.

One caveat in that ADR is worth a picture, because it is the grain the whole feature turns on: a stop is picked once, and everything derived from it comes in pairs when two lines serve it.

03-colour-means-two-things

So an interchange stop consumes two palette slots rather than one, and the caption above the figure counts stops and series separately whenever the two numbers differ.

## The one real complication: the data does not cover the same years

The chart covers 2009 → 2026, and stop data exists for twelve months, July 2025 to June 2026. So the panel has to be honest about that rather than quietly show less than you asked for. It states both date ranges before it fetches anything; where your range only partly overlaps, it names the months it actually has; and where nothing overlaps, it says so and offers a button that moves your range to a period with data.

No-overlap empty state

That button is the only thing that ever moves your date range, and it goes through the same setters a chart drag uses, so the URL, the date pickers and the chart all follow one press.

## What lands in the URL

The whole view is shareable, which means every piece of panel state is a parameter:

| Parameter | Holds |
| --- | --- |
| `stops=1` | the panel is open |
| `measure=offs\|both` | which figure the panel ranks and draws by |
| `stop=<key>,<key>` | the stops you have picked, in the order you picked them |
| `stopq=<text>` | the table's search text |

`stop=` held one key before it held several, and it still holds one perfectly well, so every link shared before this PR keeps working. A comma cannot occur inside a stop key — the charset is `^(bus|rail):[a-z0-9-]+$` — which is what makes splitting unambiguous, and it is the same property `lines=801,802` already relies on.

The order matters more than it looks: the chart takes a hue by position, so inserting anywhere but the end would recolour every series already on screen.

## Keeping the page fast

The stop data is large, so none of it downloads unless you open the panel, and none of this code sits on the first-paint path — `OutputArea` is the only importer of `useStopView`.

| File | Downloaded when |
| --- | --- |
| `stop-ridership.rail.json` (89 KB) | panel is open |
| `stop-ridership.bus.json` (5.3 MB) | panel is open **and** your range overlaps the stop data **and** you have picked a line the rail file does not cover |
| `stop_locations.json` (1.6 MB) | lazily, alongside the rail fetch |

Whether a line counts as "bus" is decided by what the rail file contains, not by the line's mode in the metadata. The G Line (901) and J Line (910) are bus rapid transit and arrive in the Bus workbook, so asking "is this a bus line?" would leave someone who picked only the G Line looking at an empty panel.

04-fetch-gate

Two e2e cases hold that branch down: one asserts a rail-only selection never requests the bus payload, the other that selecting a bus line requests it exactly once.

## One deliberate deviation from the plan

The plan said to register the map source and layer inside the existing `map.on('load')`. They are registered on first use instead — the first time there are actually circles to draw — because `e2e/map.spec.ts:217` asserts the map's layer stack is exactly `['background','lines-all','lines-selected']`, and this PR may not modify that spec.

Registering on first use keeps the assertion true and costs nothing to a reader who never opens the panel. The source is still added exactly once, guarded by `map.getLayer`, the map is never rebuilt, and every later change is a `setData`, `setFilter` or `setPaintProperty`.

## What the docs record

[ADR-0012](https://github.com/streetsforall/metro_ridership_app/blob/feat/stop-panel/docs/adr/0012-the-stop-payload-is-columnar-dictionary-encoded-and-versioned.md) records the stop wire format and [ADR-0013](https://github.com/streetsforall/metro_ridership_app/blob/feat/stop-panel/docs/adr/0013-a-stops-identity-is-its-normalised-name.md) records name-as-identity. Neither is a new decision: both were made in PRs 1 and 2 and had lived only in `scripts/README.md` and the module comments, where they read as description rather than as choices. The numbering runs out of order because those two numbers were held while the colour decision took 0014.

0013 is worth reading twice, because it marks a trap. `STATION_ORDER`'s numeric prefix looks like an id and is a per-route sequence, so in 2025-12 Union Station is 1026 on the A Line, 4001 on the B and 5001 on the D.

`docs/ROADMAP.md` row 5 is ticked, and everything here is recorded there under *What PR 5 found*.

## Gates

- `npx tsc -b`, `npm run lint` and `npm run build` — clean
- `npx vitest run` — **1160 passed, 44 files** (was 1061 across 42)
- `stop-panel.spec.ts` and `stop-map.spec.ts` in the Playwright Docker image — **41 passed**, against the committed baselines
- Main is merged down (`dd2c97f`, the pin/Month Readout fix from #226). Clean merge, no conflicts, and no baseline moved: that commit touched no snapshot PNGs and this branch changes nothing about hovering

Checked against the real committed data (A Line plus line 204, default date range): 121 table rows, 98 circles in view, radii 4.1–22.0 bus and 8.1–22.0 rail, each mode scaled against its own busiest stop.

## Tests

Two new specs and two new snapshot folders, both stubbing the payloads through `e2e/stop-fixtures.ts` with real stop keys so the GTFS join actually resolves. No pre-existing spec, helper or baseline is modified, with one sanctioned exception: eleven Linux baselines moved because the **Stop Ridership** checkbox is new chrome in the filter bar, and four more moved when the table gained a checkbox column and a search band.

`stop-panel-no-overlap` did **not** move, and that is a design choice rather than luck: the new chrome renders after the panel body's early returns, so a period with no stop data shows no search box for a table that is not there.

`stop-map.spec.ts` is named that way on purpose. The `map` project's `testMatch` and the viewport projects' `testIgnore` are both the unanchored pattern `/map\.spec\.ts/`, so the filename alone routes this spec into the `map` project — where SwiftShader and `deviceScaleFactor: 1` are pinned — and out of the other two. Renaming it would move it silently.

## Known gaps — reported, not fixed here

1. **Nothing caps the selection, and `Select All` is scoped by the search instead.** This is strict analogy with the line selector, which caps nothing either. The exposure is real: press `Select All` on an unsearched five-line table and you pick ~800 stops, and the chart will try to draw all of them. The palette cycles past eight, so colours repeat well before that.
2. **A long stop name still clips out of the legend at mobile width.** Two mitigations were tried first — the measure joins a legend label only under `both`, and the `SELECT` heading became `sr-only` so the column stopped widening the table, which brought `AVG. BOARDINGS` back into view. What remains is arithmetic: `7th Street / Metro Center Station · A Line` is 41 characters in a 294 px box, so Chart.js truncates it whatever the format. The row beneath names the stop in full, and the desktop legend is clean.
3. **The sparkline column does not mount on mobile until you scroll sideways.** 696 px of table in a 294 px scroller puts the last column outside the box, so the observer is right not to draw it. The horizontal scroll is pre-existing — six columns were already 472 px in that 294 px — though this PR widens it with the sparkline and again with the checkbox. It is left visible because the column was asked for and is reachable; hiding it below `sm` is one class.
4. **Diagram 10 is now five parameters behind.** `10-url-contract.mmd` is titled "The nine parameters" and enumerates them; this batch added `stops`, `measure`, `stop` and `stopq`. It is the same file #181 and #182 are rewriting, which is why it was left alone, and `docs/ROADMAP.md` now says so explicitly rather than leaving the staleness implicit.
5. **`src/stops/index.ts` exposes no per-stop series.** One is assembled in `src/utils/stopSeries.ts`; it would sit better beside `stopMetrics`, in a PR allowed to touch `src/stops/`.
6. **`stop_locations.json` is 1.6 MB and `stops.types.ts` calls it "small".** Worth correcting that comment.
7. **`stop=<key>` gets percent-encoded.** `URLSearchParams.toString()` escapes the `:` regardless. It decodes back to the same key, so sharing works, and the comma separating several keys survives unescaped.

## Checklist

- [x] `npm run lint`, `npm run test` and `npm run build` pass
- [x] Renamed or deleted an exported symbol? Grepped `README.md`, `CONTEXT.md`, `CLAUDE.md` and `docs/`, including `docs/architecture/mermaid/` and `captions.md`. `selectedStopKey` and `parseStopKeyParam` appeared only in `docs/ROADMAP.md`, which is updated.
- [ ] Changed a diagram source? No diagram source changed. `10-url-contract.mmd` needs one and is deliberately left to #181/#182 — see *Known gaps 4*.
- [x] Changed the UI? Screenshots above, and Linux baselines regenerated with the run restricted to `e2e/stop-panel.spec.ts`.


mle17:
## Review — #223 `feat/stop-panel`

Read `7220c46..d455149` against the plan (Part 2 §2.5–§2.7), `CLAUDE.md`'s invariants, `CONTEXT.md`,
ADRs 0001 / 0004 / 0005 / 0007 / 0009, `docs/how-it-works.md` and `docs/guides/testing.md`.

**Verdict: approve with fixes.** No blocking findings. One user-visible defect worth fixing in this
PR, two smaller ones, three nits.

---

### Gates

| Gate | Result |
|---|---|
| `npx tsc -b` | clean |
| `npm run lint` | clean |
| `npx vitest run` | **970 passed, 41 files** |
| `npm run build` | clean |

Playwright and Docker not run, per brief.

### Must-not-drift checklist

1. **No second window rule** ✅ — `isInMonthWindow` appears only in `src/stops/buildStopView.ts`
   (pre-existing, #180). Nothing new in the diff touches `startDate`/`endDate` arithmetic.
   `stopCoverage.ts` compares two axes the derivation produced; `stopSeries.ts` is handed
   `view.months`. Both say so in their own doc comments.
2. **`src/stops/` seam intact** ✅ — grep for `../stops/<internal>` outside the folder: 0 hits.
   Every consumer goes through `index.ts`.
3. **Radius / colour / filtering come from the module** ✅ — paint is `['get','radius']` /
   `['get','color']`; the measure and the selection ring use fill+stroke only. `Map.test.tsx`
   asserts `setPaintProperty` is never called with `circle-color`.
4. **`window.__metroMap` present; sources added once** ✅ — `Map.tsx:236` assign, `:401` delete.
   `map.getLayer('stops-selected')` is the guard; everything after is
   `setData` / `setFilter` / `setPaintProperty`. Unit-tested via a `getLayer` mock backed by the
   real `addLayer` record, which is the right way to write that test.
5. **5.3 MB payload off the first-paint path** ✅ — verified from the actual build, not by reading:

   ```
   dist/stop-ridership.bus.json      5,453.44 kB   ← emitted asset, not in any JS
   dist/assets/index-*.js              504.62 kB   ← entry
   dist/assets/OutputArea-*.js       1,099.12 kB   ← lazy
   dist/assets/stop_locations-*.js   1,013.81 kB   ← its own chunk, from the `import()`
   ```

   `useStopView`'s only importer is `OutputArea`. `stop_locations.json` lands in its own chunk
   rather than inflating `OutputArea`'s.
6. **URL state wired both halves** ✅ — `showStops` / `stopMeasure` / `selectedStopKey` in the lazy
   initialisers *and* in the sync effect *and* in its dep array. Non-default-only writes, matching
   the params above them.
7. **Boardings / Alightings** ✅ — table headers, popup rows, toggle labels, chart legend. `ons`/
   `offs` survive only as `StopMeasure` literals and URL values, which `CONTEXT.md` explicitly
   blesses. Two tests assert `not.toMatch(/\bons\b|\boffs\b/i)`.
8. **No existing e2e touched** ✅ — `git diff --stat -- e2e` → 10 files, 565 insertions, **0
   deletions**, every PNG `Bin 0 -> N`. `playwright.config.ts` untouched. 48 + 7 = 55, which matches
   the `testing.md` edit.
9. **Chart.js registered once** ✅ — `.register(` appears only at `src/chart/index.ts:23`.

Two extras I checked because they're cheap and would be expensive to miss:

- `stop-map.spec.ts` really does route to the `map` project — `testMatch: /map\.spec\.ts/` is
  unanchored, and the committed baseline is `stops-selected-map-linux.png`. The spec's own header
  comment warns about the rename hazard, which is the right place for it.
- `parseStopKeyParam`'s `^(bus|rail):[a-z0-9-]+$` against all **6,895** committed stop keys →
  **0 rejected**. A tighter-than-the-data regex would have silently dropped `stop=` from shared
  links.

---

## Findings

### 1 — NON-BLOCKING (fix before merge) · the panel says "No stop-level data has been ingested yet" while it is loading

`StopPanel.tsx` — `<StopCoverageNotice>` renders **outside** `body()`, unconditioned on `isLoading`:

```tsx
<div className="mt-2">
  <StopCoverageNotice state={coverageState} … />
</div>
{body()}
```

`coverageState` comes from `view.coverage`, and `buildStopView` returns
`{from: null, to: null, overlapsWindow: false}` for `records === null` — the loading state.
`stopCoverageState` maps `from === null` → `'no-data'` → *"No stop-level data has been ingested yet."*

So from panel-open until the rail payload lands, the panel reads:

> No stop-level data has been ingested yet.
> Loading stop ridership…

Same on failure: `railStatus === 'error'` leaves `records` null, so the notice claims nothing was
ever ingested while the body below it says the fetch failed. A network problem is reported as a
missing dataset.

This is precisely the ordering rule the component already states for itself —
`// Order matters: `overlapsWindow` is `false` while nothing has loaded, so the loading state has to
be answered before the empty one`. `body()` honours it; the notice above it doesn't.

Not caught, because the covering test asserts only the *other* empty state
(`StopPanel.test.tsx`, "shows loading rather than the empty state while a payload is in flight"
checks `stop-coverage-empty` is null, never `stop-coverage-no-data`), and no e2e captures the
transient. Whatever the fix, the test wants the second assertion.

### 2 — NON-BLOCKING · selecting a stop is mouse-only, and `aria-selected` on a plain `<tr>` isn't valid ARIA

`StopTable.tsx` — the row carries `onClick` with no `tabIndex`, no key handler, no `role`. The map
circle is the only other route in, and that is a mouse click too, so a keyboard user can reach the
ranked table but cannot select a stop, and the per-stop series is unreachable for them.

Separately: `role="row"` supports `aria-selected` only inside a `grid`/`treegrid`. On a plain
`<table>` it's ignored, so the assertion in `StopTable.test.tsx` ("marks the selected row") is
pinning an attribute assistive tech won't read. The visual `bg-stone-200` is doing the real work.

The sortable `<th onClick>` has the same keyboard gap, but that one matches `LineSelector.tsx:366`
exactly — pre-existing pattern, out of scope here.

### 3 — NON-BLOCKING · `how-it-works.md` now says `lines-all` is filtered, and it isn't

```
Three layers: `lines-all` (dimmed), `lines-selected` (brand colours) and `stops-selected`
(circles, above both), each filtered by the selected line ids via `setFilter`.
```

`lines-all` carries no `filter` at all (`Map.tsx:250`) — that's the whole point of it. The pre-edit
sentence scoped the clause correctly to `lines-selected`; folding three layers into one clause lost
the scope. Everything else in the docs diff checks out against the code, including the intent-gate
description, the BRT-in-the-bus-payload quirk, and the baseline counts.

### 4 — NIT · orphaned doc comment in `OutputArea.tsx`

Two JSDoc blocks stack immediately above `windowMonths`. The first ("Jump the Month Window to the
Stop Coverage Window") documents `useCoverageWindow`, declared two statements later. The second is
`windowMonths`' own. Reads as if `windowMonths` does the jumping.

### 5 — NIT (unverified, no browser run) · the shared popup probably blanks for a frame when the pointer slides off a circle onto its own route

`onMouseLeave` is registered on both `lines-selected` and `stops-selected` and calls
`popup.remove()`. A circle sits directly on top of its route, so leaving the circle while staying on
the line fires stops-`mouseleave` *and* lines-`mousemove` from the same DOM event; registration
order puts the `remove()` last. The next mousemove restores it. I couldn't confirm this without
Playwright, and it may be invisible in practice — flagging it because the two layers deliberately
share one popup instance, which is where that class of bug lives.

### 6 — NIT · the feature ships with no way to turn it on

`toggleShowStops` is exported from the hook and bound to nothing; `stops=1` is the only entry point.
`ROADMAP.md`'s "What PR 5 found" explains this well (baseline gate + #181/#182 both rewriting
`DateRangeSelector`), so this is a heads-up rather than an objection — just worth being explicit
that merging this ships a feature no reader can discover, until Panel Settings lands.

---

## What's good, specifically

- **The intent gate is asked the right question.** "Does the rail payload already serve this line",
  answered off the rail records, rather than "is this a bus line" off the metadata — which would
  have left a G Line reader staring at an empty panel. The fallback direction (unknown line → bus)
  is the safe one, and it's argued in the code rather than asserted.
- **`overlapsWindow` is read, never re-derived.** `useStopView` gates the 5.3 MB fetch on the
  rail-only view's own coverage answer. That is the single highest-leverage decision in the diff
  for the invariant this repo cares most about, and it's the reason a window in 2015 costs 89 KB
  and stops.
- **`AbortController`s in a ref, not per effect run.** The comment explains why a per-run controller
  would cancel an in-flight download every time the selection changed. Correct, and non-obvious.
- **The empty-deps init effect lost its `exhaustive-deps` exemption.** Moving `lines` to `linesRef`
  inside `load` removed a suppression rather than adding one.
- **Tests pin behaviour, not implementation.** The ones that stood out: `Map.test.tsx` keeping
  `layerIds` outside `vi.clearAllMocks()` so "nothing was re-added" can't pass for the wrong reason;
  `StopTable`'s "sinks a readout with no figures below one reporting zero" (ADR-0004 at stop grain);
  `useStopView`'s "fetches each payload once, however often the inputs change"; the round-trip test
  that asserts `params.get('stop')` rather than the literal `%3A` spelling, with the reasoning
  written down.
- **The e2e fixtures use real stop keys** so `attachStopLocations` actually joins — an invented key
  would decode fine and vanish from the map. That failure mode is documented in the fixture header.
- **Deviations from the plan are stated, not silent.** Panel below the row instead of between
  `SummaryData` and the map; `stop=` percent-encoding; `stop_locations.json` being 1.6 MB and not
  "small". Each is written down where the next reader will hit it.

---

<sub>Ran: `npx tsc -b`, `npm run lint`, `npx vitest run`, `npm run build`, plus a regex check of
`parseStopKeyParam` against all committed stop keys. No Playwright, no Docker. No files modified.</sub>

mle17:
## Review — #223

Read the diff first, formed my own view, then checked the drift list. **No blocking findings.**

`npx tsc -b` → clean · `npm run lint` → clean · `npx vitest run` → 974 passed, 41 files · `npm run build` → green.

---

### Must-not-drift checklist

| # | | |
|---|---|---|
| 1 | No restated window arithmetic | ✅ `buildStopView` is the only caller of `isInMonthWindow`; `stopCoverage.ts` compares axis-against-axis, `stopSeries.ts` maps over `view.months`. No `getFullYear`/`getMonth`/`startDate` comparison anywhere in the new components or utils. |
| 2 | `src/stops/` internals sealed | ✅ grep for `../stops/<file>` outside the folder → nothing. Same for `../ridership/`. |
| 3 | Radius / colour / filtering from the module | ✅ `'circle-radius': ['get','radius']`, `'circle-color': ['get','color']`, filter from `lineIds`. `MEASURE_PAINT` sets only opacity + stroke width, which is the encoding decision, not the scale. Pinned by *reads radius and colour off the feature rather than recomputing them* and *encodes the measure in fill and stroke rather than a second colour*. |
| 4 | `window.__metroMap` + add-once | ✅ `Map.tsx:236` / `:401` intact. `ensureStopLayer` guards on `map.getLayer`; `Map.test.tsx`'s `layerIds` array is kept outside `vi.clearAllMocks()` so the guard is answered by what the component actually added — that detail is what stops the re-add test passing for the wrong reason. |
| 5 | 5.3 MB off the first-paint path | ✅ verified against a real build: `dist/stop-ridership.bus.json` 5,453 kB **emitted asset**, `stop_locations` its own 1,013 kB async chunk, entry `index-*.js` 504 kB with zero fixture keys in it, `index.html` preloads only the entry. |
| 6 | URL state through both halves | ✅ three lazy `useState` initialisers + all three in the sync effect's dep array, written only when non-default. |
| 7 | Boardings / Alightings | ✅ every reader-facing string. `ons`/`offs` survive only as `StopMeasure` literals and wire columns, which CONTEXT.md explicitly sanctions. |
| 8 | No existing e2e touched | ✅ `git diff --stat 7220c46..origin/feat/stop-panel -- e2e` → 10 files, **all `A`**, 565 insertions, 0 deletions. |
| 9 | One Chart.js registration | ✅ `ChartJS.register` appears once, `src/chart/index.ts:23`. `StopSeriesChart` reaches it through the barrel. |

---

### Last commit (`09e2add`) — the three changes

All three are right, and the state machine still terminates in something sensible everywhere.

**`no-data` → `unknown` is the correct cut.** `buildStopView` returns `coverage: {from: null, to: null}` for `records === null`, which is loading *and* failed-fetch, so `no-data` was reading a build-time fact out of a runtime one. Moving it to `minMonth === null` in the notice is the right home — the manifest is filled from the files at build time, so it is right before the first fetch and right when every fetch fails.

**Every state still renders.** Walked all of them:

- `unknown` + loading → span line + "Loading stop ridership…"
- `unknown` + failed → span line + "could not be loaded" (no longer "never ingested" — the regression the commit names)
- `unknown` + settled, empty payload → span line + "No stop-level data for the selected lines in this period." Reachable, and reads correctly.
- `no-overlap` → span + button, body `null`
- `partial` / `full` → label or nothing, plus table

Removing `no-data` from `StopPanel`'s early-`null` did not orphan anything: `unknown` now falls through to the readout-count branch, which is the sentence you'd want.

**Non-null assertions on the button are safe.** `stopCoverageState` returns `'unknown'` before `'no-overlap'`, so `coverage.from!`/`to!` are unreachable-when-null by construction.

**`aria-selected` → `aria-current`** is correct — `aria-selected` is only mapped on a row inside `grid`/`treegrid`. And the negative test (*leaves an unselected row unmarked rather than marking it false*) pins the absent-attribute behaviour, not the implementation.

**Tests pin behaviour, not shape.** `it('leaves other keys alone')` is the one I'd have asked for. The one I'd push back on is in `StopCoverageNotice.test.tsx`: the suite now has *no* coverage of the `minMonth === null` branch, and the doc comment says so honestly ("not reachable in this suite") — but the plugin explicitly supports absent payloads as a state, so that branch is live in a fresh clone before the first ingest. It is untestable only because the manifest is a module-level import; `vi.mock('virtual:stop-ridership-manifest')` would reach it.

---

### Findings

**NON-BLOCKING — 1. The bus fetch blanks a panel that already had data.**
`isLoading` is true while the 5.3 MB bus payload downloads, and `StopPanel.body()` answers `isLoading` second, before anything else that could render. So: reader selects a rail line, gets the table, selects a stop, gets the series — then adds a bus line, and both vanish behind "Loading stop ridership…" for the length of a multi-megabyte download. The rail data it already has is exactly what they were looking at. A "loading more…" note beside the existing table keeps it; the loading-wins ordering is right for the *first* payload and wrong for the second.

**NON-BLOCKING — 2. A partial failure is reported as a total one.**
`hasFailed = railStatus === 'error' || busStatus === 'error'`, and `hasFailed` is the first branch in the body. Rail lands, bus 404s → the whole panel becomes "Stop-level ridership could not be loaded." and the rail readouts that *did* load are discarded. Same shape as (1) and the same fix: the two payloads have independent fates and the panel collapses them into one.

**NON-BLOCKING — 3. The coverage button's label and its action come from different sources.**
Label is `coveredSpan` — `minMonth`/`maxMonth` off the manifest, the union of both payloads, build time. The click sends `coverage.from`/`coverage.to`, which in the `no-overlap` state is the **rail-only** span, because the bus fetch is gated on `overlapsWindow` and so has not happened.

They agree on today's data — I checked both payloads, rail and bus are each `2025-07 … 2026-06`, 12 months — so this is latent, not live. But it is *already visibly wrong under the e2e stub*, which is the cheapest possible demonstration: `stop-panel-no-overlap.png` renders "Show Jul 2025 – Jun 2026" from the real manifest while `#use-stop-coverage-window` sets Jul–Dec 2025 from the rail fixture, and the spec asserts the December end (`#end-month` = `'11'`). Two sources, two answers, one button, and a committed baseline showing them disagree.

The manifest is the authority on "where the data is", which is the button's whole purpose — passing `minMonth`/`maxMonth` to the click as well as to the label would collapse it to one source.

**NON-BLOCKING — 4. A focusable row is not an announced one.**
`<tr tabIndex={0} onClick onKeyDown>` with no `role="button"`. A screen-reader user tabs onto it and is told it is a row; nothing says Enter does anything. Two consequences worth weighing together: the missing role, and ~800 tab stops for a five-line selection sitting between the table and whatever follows it. `LineSelector` solves the same problem by putting a real `<button>` in the cell — the accessible name comes free and the tab order stays proportionate. Worth following rather than inventing a second pattern.

**NIT — 5. Sortable headers are still mouse-only.**
`<th onClick>` with `cursor-pointer` and a correct `aria-sort`, but no `tabIndex`/`onKeyDown`. This matches `LineSelector.tsx:366` exactly, so it is consistent rather than a regression — flagging only because the same commit made rows keyboard-reachable and left the headers behind, in the same file.

**NIT — 6. One stop on two lines yields two rows with the same `data-qa`.**
`key` is `${line_name}-${key}`; `data-qa` is `stop-row-${key}`. Both rows take `aria-current="true"` when that stop is selected, while only the first readout's series is drawn. `StopPanel` documents "the first in readout order wins and the caption names its line" — the highlight doesn't say so, and it also makes `[data-qa="stop-row-…"]` ambiguous for any future e2e selector.

**NIT — 7. `setFilter('stops-selected', …)` is not load-bearing.**
`buildStopView` already filtered markers to `lineIds`, so the layer filter can never exclude a feature. It matches the plan and it is free — just noting it is belt-and-braces, not the mechanism.

**NIT — 8. `buildStopPopupHTML` interpolates `name`/`lineName` unescaped.**
Identical to the existing `buildPopupHTML`, and both read pipeline-generated data, so not exploitable. Copied pattern, mentioning for the record only.

---

### Things I checked and liked

- **`useStopView`'s controller ref.** Holding the `AbortController`s for the hook's lifetime rather than per-effect-run is the right call and the comment says why: a per-run controller would cancel an in-flight download every time the selection changed. I also walked React 19 StrictMode's double-invoke — the `[]` cleanup aborts, then the second setup sees the same `'idle'` closure and restarts, so it settles on one live fetch rather than hanging.
- **The bus gate asks the right question.** "Does the rail payload already serve this line", read off the rail records, instead of "is this a bus line" off metadata — which would have left a G Line-only selection staring at an empty panel. Failure direction is safe (a line in neither payload falls to the bus side).
- **`stop-map.spec.ts`'s name.** The `map` project's `testMatch` and the viewport projects' `testIgnore` are both the unanchored `/map\.spec\.ts/`, so the name is load-bearing and `testing.md` now says so in the place someone renaming it would look.
- **Fixtures use real stop keys** because `attachStopLocations` joins against bundled geometry — an invented key decodes fine, appears in the table, and is silently missing from the map. That is exactly the failure a fixture is supposed to not have.
- **`queryRenderedFeatures` over pixel counts.** A failing stop-key list localises a broken filter far faster than a diff image, and the screenshot is left to cover what a key list can't (radius, colour, stacking).
- **Assert-then-shoot** in `stop panel — the ranked table is the primary readout`: the vocabulary and the row keys are checked before the capture, so a mistyped param can't be baked into a green baseline.
- **`ROADMAP.md`'s "What PR 5 found"** is honest about the four things this PR chose not to do, including the missing visibility control — `toggleShowStops` is bound to nothing, the panel ships URL-only, and the reason (any filter-bar chrome moves `visual.spec.ts`'s six, which this PR's own gate forbids; #181/#182 are rewriting that area anyway) is a real one with a named follow-up. Agreed with the call. It does mean the only exercise `toggleShowStops` gets is its own test.
- Docs match the code, including the `lines-all`-is-unfiltered correction and the `stop_locations.json`-is-1.6 MB admission against `stops.types.ts` calling it "small".

---

**Verdict: approve.** Nothing here blocks the merge. (1)–(3) are worth a follow-up commit — (3) especially, because the disagreement is already committed as a baseline.

mle17:
Closing in favour of a four-PR stack. Same work, same commits' worth of content — split so each piece can be read properly rather than skimmed at +7,460/68 files.

| | | |
|---|---|---|
| #231 | the panel, the ranked table, URL state, the coverage notice | ← `main` |
| #232 | the ridership-over-time column | ← #231 |
| #233 | the stop series chart and its colour rule (ADR-0014) | ← #232 |
| #234 | the map's circle layer | ← #233 |

Nothing was dropped in the split. With all four stacked, `git diff feat/stop-map-circles feat/stop-panel` is empty apart from the ROADMAP row that describes the split itself.

Both review passes on this PR still apply, and their three findings were already fixed on this branch before the carve — `1065e26` answered the two payload-fate findings eight minutes after the second review was posted, and the coverage button's label and action now both read `minMonth`/`maxMonth` off the build-time manifest.

Each slice was gated on its own: tsc, lint, build, vitest and the full Playwright suite in `mcr.microsoft.com/playwright:v1.62.1-noble`. Only #231 moves a pre-existing baseline.

